PR #119, opened 2013-03-17
ignore multiple rules + ~ syntax

grosser:
@acrmp


ebrahim-moshaya:
I got this warning:
FC015: Consider converting definition to a LWRP: ./definitions/default.rb:1
Adding "# ~FC015" to the first line in the definitions/default.rb file did not ignore the warning...

PR #169, opened 2013-09-05
Adding support for additional / custom reporters

danleyden:
- bumped version
- Refactored a few things to make the report outputting more common.
- New concept of reporter
  - previously had ContextOutput and SummaryOutput
  - these have been separated out as individual reporters
  - additional CheckstyleOutput reporter created
- Added extra (optional) command line params
  - require
    allow specifying reporters from outside
  - report-dest
    passed in to reporters through a call to destination attribute and
    expected to be used to specify e.g. output file location
  - reporter
    takes the class name (qualified) of the reporter to use
    e.g. FoodCritic::ContextOutput, FoodCritic::CheckstyleOutput
- Added source attribute to rule DSL, automatically populated as the source
  file/line of the rule block, providing a common attribute to look at to
  obtain that info
- Altered the rake task to allow propogation of options into the reporter

(Sorry about the big pull)


lamont-granquist:
Seems like a good idea but probably way out of date with the code by now.

clintoncwolfe:
Refs #202

clintoncwolfe:
For those looking for something now, in 2013 I had a stream-based foodcritic-to-junit parser, which is probably iffy for several reasons: https://github.com/clintoncwolfe/chef-ci-tools/blob/master/bin/foodcritic2junit.pl

Several of us at Chef Summit Hack Day 2015 are looking into working on getting CI-consumable output from chef analysis and testing tools - see https://e.chef.io/p/HackDay-CI-reporting-test-formats

lamont-granquist:
So one thing that @coderanger talked to me about was simply porting foodcritic rules to rubocop plugins.  That'd get us away from dumping the ruby AST into XML and using nokogiri and XPath to parse it all.  It'd also leverage us a lot of work being done on rubocop with the ability to have different warning levels, and disabled-by-default, and i'd expect they're much further along with output plugins (although i haven't checked).

clintoncwolfe:
FWIW, I went ahead and tackled the conflicts - it wasn't bad at all.  See https://github.com/clintoncwolfe/foodcritic/tree/ccw/reporter-merge .

There are things I would suggest doing differently, even assuming it's a foodcritic-exposes-plugins approach - allow multiple output streams, call them "formatters" instead of "reporters", etc.  I'm working on that here https://github.com/clintoncwolfe/foodcritic/tree/ccw/formatters .   I'm generally following the approach taken by rubocop's formatters.

In both cases, I was on a plane, and could not get the testing gems I needed, so no TDD on this yet; not ready for primetime.

As for re-implementing foodcritic as a series of rubocop plugins: that sounds great, and I have no issue with that down the line.   For the near-term, I want pluggable output formats, and I don't need a full rewrite to get it.

tas50:
Closing this out since it'll need a significant rebase at this point. If it was rebased and the tests pass, I'd be down to get this included.

PR #182, opened 2013-10-22
Update dsl, add RuleDsl#update

jasonLaster:
It would be helpful to be able to update an existing foodcritic rule, either to change the behavior or to whitelist one of the checks. 

```ruby
update "FC041" do
  recipe do |ast|
  end
end
```


odcinek:
I prefer to add new & disable old rule, just to avoid potential confusion. But yeah, this might be useful - @jasonLaster you want to roll the functionality into `rule` keyword and provide some tests?

PR #203, opened 2014-01-16
The rake task needs to require foodcritic.

docwhat:
This makes using the LintTask easier:

```
require 'foodcritic/rake_task'

FoodCritic::Rake::LintTask.new
```


odcinek:
The actual usage is
```
require 'foodcritic'
task :default => [:foodcritic]
FoodCritic::Rake::LintTask.new
```
but whatevs

PR #205, opened 2014-01-16
Add task for build_features (aruba features)

docwhat:
Add a task to run the build_features (@build and @context).

It took me a while to figure out how to run those so I could create
the test(s) for #204.

It may be worthwhile adding these to the default task as well.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -28,6 +28,14 @@ Cucumber::Rake::Task.new(:features) do |t|
   t.cucumber_opts += ['features']
 end
 
+desc 'Run Cucumber build features via Aruba'
+Cucumber::Rake::Task.new(:build_features) do |t|
+  t.cucumber_opts = ['-f', 'progress', '--strict']
+  ENV['FC_FORK_PROCESS'] = true.to_s
+  t.cucumber_opts += ['-t', '@build,@context']
+  t.cucumber_opts += ['features']
+end
+
 desc 'Run RuboCop'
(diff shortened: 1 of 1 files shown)


lamont-granquist:
looks like this needs a rebase

docwhat:
Oh wow. I don't even remember this. Are you still interested?

lamont-granquist:
Heh, I'm not even sure what it is, but seems fine to add a target so that nobody else has to re-do the work you did to figure it out...

acrmp:
Thanks Christian.

I'm not sure this is going to do what we want though because the assignment to `ENV['FC_FORK_PROCESS']` will happen when loading the Rakefile, not only when executing that particular task.

A bit more background: The `FC_FORK_PROCESS` env var is to allow you to run the foodcritic features by shelling to foodcritic each time - this significantly slows the build but does catch another class of errors. I typically used to run the build with `FC_FORK_PROCESS=true` only before cutting a new release.

Rakefile
```diff
@@ -35,6 +43,6 @@ RuboCop::RakeTask.new(:rubocop) do |task|
 end
 
 desc 'Build the manpage'
-task :man do
+task(:man) do
```
philoserf: Why should this style change be part of the pull request?

docwhat:
Because it wouldn't work without it. Ruby itself complained. 

I've never gotten that style of function-argument-block to work without parentheses.

tas50:
I'm closing this out since we're in the process of migrating away cucumber testing

PR #229, opened 2014-04-23
Update linter.rb

chazzly:
pull in and combine multiple .foodcritic files:
the original location in each cookbook, ~/.foodcritic , ~/.chef/.foodcritic

This allows for global settings - ex: I don't agree with FC###, and don't want to check for it EVER.  add "~FC###" to ~/.chef/.foodcritic and Done!


chazzly:
Moved to PR #290

PR #231, opened 2014-04-29
FC007: accept 'suggests' and 'recommends'

matthiasr:
Chef supports the `suggests` and `recommends` as keywords for soft dependencies. They are a NOOP and the semantics are a bit [unclear](http://tickets.opscode.com/browse/CHEF-3870) but they're [used](socrata-cookbooks/java@90ecd3db) for "most people don't need this" dependencies, e.g. depending on the `windows` cookbook.

Fixes acrmp/foodcritic#159.

### DIFF
--- a/lib/foodcritic/api.rb
+++ b/lib/foodcritic/api.rb
@@ -97,14 +97,14 @@ module FoodCritic
       File.basename(file)
     end
 
-    # The dependencies declared in cookbook metadata.
+    # The dependencies, recommendations and suggestions declared in cookbook metadata.
     def declared_dependencies(ast)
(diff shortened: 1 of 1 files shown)


lamont-granquist:
You probably don't want to be using `suggests` or `recommends` at all since if we ever do implement them then if the semantics we implement do not agree with what you intended, your cookbook will be broken.

odcinek:
FC052 and FC053 mark those as unimplemented, https://github.com/acrmp/foodcritic/commit/aa554ce2b8b64a2ddb122981a30126431770612b

PR #243, opened 2014-06-04
make the --epic-fail option take special 'all' value

aerickson:
Adds the ability for the --epic-fail option to take 'all' value. It will make all rules 'epic fail' when they fail.


odcinek:
`--epic-fail` already takes `any` to fail on all warnings

PR #248, opened 2014-07-15
New cookbook_maintainer api methods

miguelcnf:
I've had the need to programatically retrieve the maintainer and maintainer_email fields from the metadata file using the FoodCritic::Api module so I've extended this locally but I think they'd be a good fit to have upstream.

Summing up what I've done:
 * created the `def metadata_field(file, field)` method to re-utilize the logic of retrieving a field from the metadata ast on multiple api methods
 * added mock tests for the new methods
 * added a mock test for the `def cookbook_name(file)`to validate it can retrieve the name of cookbook from the metadata file - it was only returning the name of the cookbook based on the file


tas50:
Thanks @miguelcnf

PR #257, opened 2014-08-07
Do not return matches on excluded files.

### LINKED ISSUE #255: FC019 ignores exclude_paths
The first thing FC019 does is reselect all of the .rb files in the cookbook directory:

```
    asts = {}; files = Dir["#{cookbook_dir}/*/*.rb"].reject do |file|
```

Shouldn't it only operation on the file that is being currently tested? This makes it impossible to exclude a path from getting this rule applied.

bytheway:
Adds another method to remove any matches that have been
made on excluded paths.

Also, changes the files_to_process to expand_paths to
return both the files we want to match, as well as all
files detected as excluded.

Fixes #255

### DIFF
--- a/lib/foodcritic/linter.rb
+++ b/lib/foodcritic/linter.rb
@@ -51,9 +51,10 @@ module FoodCritic
       warnings = []; last_dir = nil; matched_rule_tags = Set.new
       load_rules
       paths = specified_paths!(options)
+      files = expand_paths(paths)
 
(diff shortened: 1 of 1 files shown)


odcinek:
I don't get the point of this. `files_to_process` is already aware of exclusions, why move this logic around? If you want to provide test cases and rebase on master I'm happy to revisit this.

lamont-granquist:
No updates in nearly a year, closing.

Issue #265, opened 2014-09-21
files_to_process doesn't include child paths under templates/default/ 


ruizink:
This piece of code doesn't scan *.erb files, other than the ones that seat under templates/default/ directory:

```ruby
template do |ast, filename|
  puts filename
end
```

I'll do a pull request with the fix right away.

Cheers.

odcinek:
I was unable to reproduce on master, and the source looks right. Closing.

odcinek:
Er, this needs correction - recursively scanning under `templates/default/` would be wrong, as according to resource docs `source` is *relative*.

PR #266, opened 2014-09-21
Fix for issue #265.

ruizink:
This should find all *.erb files throughout the folder tree under templates.


odcinek:
https://github.com/acrmp/foodcritic/issues/265 is closed, this is no longer needed.

PR #270, opened 2014-10-01
Add data bag JSON validation rule

### LINKED ISSUE #269: New check: validate data bag JSON
Have a check that validates that all JSON files in a cookbook's data bag directory is valid. Right now I do it in a Rakefile, but it should be pretty simple to adopt:

```
databag_items.each do |file|
  begin
    contents = File.open(file, 'rb').read
    bag = JSON.parse(contents)
  rescue JSON::ParserError
    raise "#{file} not valid JSON!"
  end
  if bag.fetch('id') != File.basename(file, '.json')
    fail "#{file} has mismatched 'id' attribute!"
  end
end
```

john-karp:
Resolve #269

### DIFF
--- a/lib/foodcritic/rules.rb
+++ b/lib/foodcritic/rules.rb
@@ -739,3 +739,19 @@ rule 'FC051', 'Template partials loop indefinitely' do
     end.map { |t| file_match(t) }
   end
 end
+
+rule 'FC052', 'Invalid data bag JSON' do
+  require 'JSON'
+  tags %w(correctness files)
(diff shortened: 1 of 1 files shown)


odcinek:
I like it. Could you rebase on master, bump the rule number and provide some tests? Also, this could be expanded to all `json`, not only databags.

tas50:
I'm going to close this one since we haven't heard back from the author in over 2 years. @john-karp If you're still interested this would be a great rule with a test

PR #277, opened 2014-11-07
add compatibility for Gherkin versions < 2.11.7

ktdreyer:
Gherkin versions prior to 2.11.7 did not have the `evaluate()` method. For backwards compatibility, provide `evaluate()` if it is not available.

### DIFF
--- a/lib/foodcritic/domain.rb
+++ b/lib/foodcritic/domain.rb
@@ -92,3 +92,13 @@ module FoodCritic
     end
   end
 end
+
+# Gherkin versions prior to 2.11.7 did not have the evaluate() method. For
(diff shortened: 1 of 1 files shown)


odcinek:
I'd rather depend on `~> 2.11.7` in `gemspec`

odcinek:
Not need to pin it anymore, since
```
Bundler could not find compatible versions for gem "gherkin":
  In Gemfile:
    foodcritic (>= 0) ruby depends on
      gherkin (~> 2.11.7) ruby

    cucumber (~> 1.3) ruby depends on
      gherkin (~> 2.12.0) ruby
```

PR #282, opened 2014-11-15
FC054, Name should match cookbook dir name in metadata

odcinek:
In Chef 12 name metadata attribute needs to match cookbook dir name. This commit makes FC045 more strict, by not only requiring name to be present, but also match.

While this might break back compat, it actually enforces what new releases require, so I guess it's ok. Spawning a new rule just for this seemed like too much.

Closes https://github.com/acrmp/foodcritic/issues/242

### DIFF
--- a/features/054_name_should_match_cookbook_dir_name_in_metadata.feature
+++ b/features/054_name_should_match_cookbook_dir_name_in_metadata.feature
@@ -1,0 +1,15 @@
+Feature: Name should match cookbook dir name in metadata
+
+  In order to avoid complications where a cookbook repository name differs from the cookbook name
+  As a developer
+  I want to set the cookbook name within the metadata
+
+  Scenario: Name mismatched in metadata
+    Given a cookbook with metadata that includes a mismatched cookbook name
+     When I check the cookbook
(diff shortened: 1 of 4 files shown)


odcinek:
Build won't fail after https://github.com/acrmp/foodcritic/pull/315 lands

coderanger:
Am I correct in understanding that this is checking that the folder the whole cookbook is in matches the name?

jaymzh:
Yup. The thing Chef used to pull the name from must match the thing Chef now pulls the name from.

coderanger:
Might want to improve the error message a bit, it wasn't super clear until I read the code and I wouldn't have assumed a command run as `foodcritic .` was applying a lint rule to a folder outside the target. Maybe just make it more verbose? `"Cookbook directory (#{dir}) does not match metadata name (#{name})."`

chazzly:
Looks like this is related to #322.  It appears to be trying to match the cookbook name to "." rather than the name of the directory.

jaymzh:
Better yet we should resolve the name of the directory properly. It's easy to figure that out. `File.realpath()` :)

PR #285, opened 2014-11-26
Features/list rules

### LINKED ISSUE #284: Add CLI option to list rules in effect
Several other linting tools provide a command-line mode that processes any config files and rule-selection options, then outputs a listing of the rules that would be enabled, then exits (without running the linter).

This sort of feature is very valuable for CI integration, because it makes it much easier to enumerate the rules that are being tested.

clintoncwolfe:
Aiming to resolve #284 

This respects tags, include_rules, and rules from gems, but does not respect per-cookbook rule configurations.


odcinek:
Neat

PR #290, opened 2014-12-19
Centralized config files

chazzly:
I've taken a crack at using yml to create a config file.

I'm not sure this is done yet, but it's most of the way there and wanted to run it by you and get your thoughts / suggestions

This would supersede #272  & #229

### DIFF
--- a/lib/foodcritic/command_line.rb
+++ b/lib/foodcritic/command_line.rb
@@ -128,7 +128,7 @@ module FoodCritic
     def get_config(cbs)
       require "app_conf"
       files_to_load = []
-      if ENV['HOME'] && File.file?(File.expand_path("~/.chef/foodcritic.yml"))
+      if ENV["HOME"] && File.file?(File.expand_path("~/.chef/foodcritic.yml"))
         files_to_load << File.expand_path("~/.chef/foodcritic.yml", ".")
       end
       [".", cbs].flatten.each do |mypth|
--- a/lib/foodcritic/command_line.rb
+++ b/lib/foodcritic/command_line.rb
@@ -93,17 +93,17 @@ module FoodCritic
 
       get_config(cbs)
 
-      @options.merge!(fail_tags: @fc_config["fail_tags"].flatten.uniq) if @fc_config["fail_tags"]
-      @options.merge!(tags: @fc_config["tags"].flatten.uniq) if @fc_config["tags"]
-      @options.merge!(include_rules: @fc_config["include_rules"].flatten.uniq) if @fc_config["include_rules"]
-      @options.merge!(role_paths: @fc_config["role_paths"].flatten.uniq) if @fc_config["role_paths"]
-      @options.merge!(environment_paths: @fc_config["environment_paths"].flatten.uniq) if @fc_config["environment_paths"]
-      @options.merge!(exclude_paths: @fc_config["exclude_paths"].flatten.uniq) if @fc_config["exclude_paths"]
-      @options.merge!(cookbook_paths: @fc_config["cookbook_paths"].flatten.uniq) if @fc_config["cookbook_paths"]
-      @options.merge!(search_grammar: @fc_config["search_grammar"]) if @fc_config["search_grammar"]
-      @options.merge!(chef_version: @fc_config["chef_version"]) if @fc_config["chef_version"]
-      @options.merge!(context: @fc_config["context"]) if @fc_config["context"]
-      @options.merge!(search_gems: @fc_config["search_gems"]) if @fc_config["search_gems"]
+      @options[:fail_tags] = @fc_config["fail_tags"].flatten.uniq if @fc_config["fail_tags"]
+      @options[:tags] = @fc_config["tags"].flatten.uniq if @fc_config["tags"]
+      @options[:include_rules] = @fc_config["include_rules"].flatten.uniq if @fc_config["include_rules"]
+      @options[:role_paths] = @fc_config["role_paths"].flatten.uniq if @fc_config["role_paths"]
+      @options[:environment_paths] = @fc_config["environment_paths"].flatten.uniq if @fc_config["environment_paths"]
+      @options[:exclude_paths] = @fc_config["exclude_paths"].flatten.uniq if @fc_config["exclude_paths"]
+      @options[:cookbook_paths] = @fc_config["cookbook_paths"].flatten.uniq if @fc_config["cookbook_paths"]
+      @options[:search_grammar] = @fc_config["search_grammar"] if @fc_config["search_grammar"]
+      @options[:chef_version] = @fc_config["chef_version"] if @fc_config["chef_version"]
+      @options[:context] = @fc_config["context"] if @fc_config["context"]
+      @options[:search_gems] = @fc_config["search_gems"] if @fc_config["search_gems"]
 
       # -v is not implemented but OptionParser gives the Foodcritic's version
       # if that flag is passed
@@ -133,7 +133,7 @@ module FoodCritic
         tmp_path = File.expand_path(mypth, ".")
         until File.basename(tmp_path) == "/"
           files_to_load << "#{tmp_path}/.chef/foodcritic.yml" if File.file?("#{tmp_path}/.chef/foodcritic.yml")
-          tmp_path = File.expand_path("#{tmp_path}/..",".")
+          tmp_path = File.expand_path("#{tmp_path}/..", ".")
         end
       end
       if @original_args.include?("--config")
@@ -161,7 +161,7 @@ module FoodCritic
             v2 || v1
           elsif v1.class.name == "Array"
             v1 << v2 if v2
-          else 
+          else
             v1.each_line.to_a << v2 if v2
           end
         end
@@ -173,7 +173,7 @@ module FoodCritic
         elsif %w{search_gems context}.include?(k) && !@fc_config[k].class.name == "TrueClass" && !@fc_config[k].class.name == "FalseClass"
           @fc_config[k] = true if @fc_config[k]
         elsif %w{tags fail_tags include_rules role_paths environment_paths exclude_paths cookbook_paths}.include?(k)
-          @fc_config[k] =  Array(@fc_config[k])
+          @fc_config[k] = Array(@fc_config[k])
         end
       end
 
@@ -220,7 +220,7 @@ module FoodCritic
     # @return [Boolean] True if the paths exist.
     def valid_paths?
       paths = options[:cookbook_paths] + options[:role_paths] +
-              options[:environment_paths]
+        options[:environment_paths]
       paths.flatten!
       if paths.any?
         tst = true
--- a/lib/foodcritic/command_line.rb
+++ b/lib/foodcritic/command_line.rb
@@ -128,7 +128,9 @@ module FoodCritic
     def get_config(cbs)
       require "app_conf"
       files_to_load = []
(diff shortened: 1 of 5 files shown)


odcinek:
I applied this locally but the cause is not yet apparent to me. I'll take a closer look later. Also, could you rebase on master and squash the commits?

lib/foodcritic/command_line.rb
```diff
@@ -1,10 +1,12 @@
-module FoodCritic
-  # Command line parsing.
+module FoodCritic # Command line parsing.
```
jaymzh: I imagine this was an errant shift-J in vim...

lib/foodcritic/command_line.rb
```diff
@@ -64,14 +67,35 @@ def initialize(args)
...
+                'Path to foodcritic.yml file to use') do |f|
+          @options[:config] = f
         end
       end
+
+      cbs = @parser.permute(args)
+
+      get_config(cbs)
+
+      @options.merge!(fail_tags: @fc_config['fail_tags'].flatten.uniq) if @fc_config['fail_tags']
```
jaymzh: can't se just do `@options.merge!(@fc_config)`?

lib/foodcritic/command_line.rb
```diff
@@ -83,6 +107,68 @@ def initialize(args)
           e.recover args
         end
       end
+
+      [:cookbook_paths, :role_paths, :environment_paths, :include_rules, :exclude_paths].each do | pth|
```
jaymzh: s/| pth|/|pth|/

lib/foodcritic/command_line.rb
```diff
@@ -83,6 +107,68 @@ def initialize(args)
...
+        @options[pth].map! { |c| File.expand_path(c, '.') } if @options[pth] && @options[pth] != [nil]
```
jaymzh: fwiw, `.` is the default second arguement for `expand_path`

lib/foodcritic/command_line.rb
```diff
@@ -83,6 +107,68 @@ def initialize(args)
...
+        @options[pth].uniq!
+      end
+    end
+
+    # Search for and load config files
+    def get_config(cbs)
+      require 'app_conf'
+      files_to_load = []
+      files_to_load << File.expand_path('~/.chef/foodcritic.yml', '.') if File.file?(File.expand_path('~/.chef/foodcritic.yml'))
+      ['.', cbs].flatten!.each do |mypth|
```
jaymzh: s/!//

lib/foodcritic/command_line.rb
```diff
@@ -83,6 +107,68 @@ def initialize(args)
...
+        tmp_path = File.expand_path(mypth, '.')
+        until File.basename(tmp_path) == '/'
+          files_to_load << "#{tmp_path}/.chef/foodcritic.yml" if File.file?("#{tmp_path}/.chef/foodcritic.yml")
+          tmp_path = File.expand_path(tmp_path + '/..','.')
```
jaymzh: prefer string interpolation to concat

lib/foodcritic/command_line.rb
jaymzh: the idea here is to allow a .chef/foodcritic.yml in any cookbook?

odcinek:
Could you add a test in `spec/foodcritic/command_line_spec.rb`? There is plenty of logic in `get_config` that I'd like to see covered

lib/foodcritic/command_line.rb
```diff
@@ -83,6 +107,68 @@ def initialize(args)
...
+        end
+      end
+      if @original_args.include?('--config')
+        ind = @original_args.index('--config') + 1
```
jaymzh: Why are you manhandling this array.... why aren't you just using options[:config] ?

lib/foodcritic/command_line.rb
```diff
@@ -83,6 +107,68 @@ def initialize(args)
...
+
+      files_to_load.flatten!
+      files_to_load.uniq!
+
+      @fc_config = {}
+      tmp_config = AppConf.new
+      files_to_load.each do |load_path|
+        next unless File.exist?(load_path)
+        tmp_config.load(load_path)
+        @fc_config.merge!(tmp_config.to_hash)do |key, v1, v2 |
```
jaymzh: nuke random trailign space

lib/foodcritic/linter.rb
```diff
@@ -123,10 +128,25 @@ def load_rules
     end
 
     def load_rules!(options)
+      rl_files = []
       rule_files = [File.join(File.dirname(__FILE__), 'rules.rb')]
       rule_files << options[:include_rules]
       rule_files << rule_files_in_gems if options[:search_gems]
-      @rules = RuleDsl.load(rule_files.flatten.compact, chef_version)
+      rule_files.delete_if {|x| x == nil}
```
jaymzh: rule_files.compact

lib/foodcritic/linter.rb
```diff
@@ -182,9 +202,9 @@ def active_rules(tags)
     def cookbook_dir(file)
       Pathname.new(File.join(File.dirname(file),
                              case File.basename(file)
-                             when 'metadata.rb' then ''
-                             when /\.erb$/ then '../..'
-                             else '..'
+                               when 'metadata.rb' then ''
```
jaymzh: this was correct before.

lib/foodcritic/command_line.rb
```diff
@@ -118,8 +204,18 @@ def version
     # @return [Boolean] True if the paths exist.
     def valid_paths?
       paths = options[:cookbook_paths] + options[:role_paths] +
-        options[:environment_paths]
-      paths.any? && paths.all? { |path| File.exist?(path) }
+              options[:environment_paths]
+      paths.flatten!
```
jaymzh: I"m not sure I understand the change here... other than doing `all?` manually...

lib/foodcritic/command_line.rb
chazzly: yes

lib/foodcritic/command_line.rb
chazzly: I started with that, but keep getting errors if the values didn't exist in fc_config.  Plus, some are arrays which needed to be flattened, uniq'd, but not all.

lib/foodcritic/command_line.rb
chazzly: Not my intention, per se, but to start in both the current directory & the specified cookbook directory and search upwards for any ".chef/foodcritic.yml" files.

Yes, technically that would allow you to put ".chef/foodcritic.yml" into each cookbook.

lib/foodcritic/command_line.rb
chazzly: I wanted a way to see which path was failing - maybe there's a better way??

chazzly:
I've rebased and made many of the changes @jaymzh mentioned.

I'm working on setting up the spec tests as requested and re-looking at line 131 (I don't remember why I did it that way, but will try re-working it).

lib/foodcritic/command_line.rb
chazzly: Ok, I did it this way because I originally had get_config running before the OptionParser to ensure the command line took precedence.   As things got re-arranged, I never went back and re-worked this.  I'll fix this as well.

chazzly:
Wow, I fixed the errors with  :include_rules and  :search_gems, everything passed on my machine - no clue what's going on.  I'll keep digging.

chazzly:
I removed some unnecessary file expansion, and that looks to have resolved most of the feature tests.

I can't seem to run the feature tests correctly in my environment, but I'll do my best to not re-submit the too many times.

chazzly:
Ok, while this did fail, it's only down to two failures in the features tests.  Both are coming from the "When I run it on the command line with the help option" test, but i can't figure out why the are failing.  When I install and run this on my system, 'foodcritic --help' shows the help message just as one would expect.  No clue why those two are failing.

lib/foodcritic/command_line.rb
```diff
@@ -77,20 +81,108 @@ def initialize(args)
...
         begin
+        cbs = @parser.permute(args)
+        rescue OptionParser::InvalidOption => e
+          puts e.message
+        end
+
+        get_config(cbs)
+
+	# Tags handled differently to allow for proper precidence between config files, cookobook files, and command line.
+        @options[:tags]= @fc_config['tags'].flatten.uniq if @fc_config['tags']
```
jaymzh: You should be able to do a sane single merge:

```
irb(main):001:0> f = {'a' => 'x', 'b' => 'y', 'c' => 'z'}
=> {"a"=>"x", "b"=>"y", "c"=>"z"}
irb(main):002:0> g = {'p' => 'l', 'q' => 'm', 'r' => 'n'}
=> {"p"=>"l", "q"=>"m", "r"=>"n"}
irb(main):003:0> f.merge(g)
=> {"a"=>"x", "b"=>"y", "c"=>"z", "p"=>"l", "q"=>"m", "r"=>"n"}
irb(main):004:0> f.merge!(g)
=> {"a"=>"x", "b"=>"y", "c"=>"z", "p"=>"l", "q"=>"m", "r"=>"n"}
irb(main):005:0> 
```

Regardless of whether there is overlap. Then walk through and flatten what needs to be flattened.

lib/foodcritic/command_line.rb
```diff
@@ -77,20 +81,108 @@ def initialize(args)
...
+          else 
+            v1.each_line.to_a << v2 if v2
+          end
+        end
+      end
+
+      @fc_config.keys.each do |k|
+        if %w(chef_version config search_grammar).include? k && !@fc_config[k].class.name == 'String'
+          @fc_config[k] =  @fc_config[k].to_s
+        elsif %w(search_gems context).include? k && !@fc_config[k].class.name == 'TrueClass' && !@fc_config[k].class.name == 'FalseClass'
```
jaymzh: `.. && ![TrueClass, FalseClass].include?(@fc_config[k].class)`

lib/foodcritic/command_line.rb
```diff
@@ -132,9 +224,20 @@ def version
...
+      tst = false
       paths = options[:cookbook_paths] + options[:role_paths] +
-        options[:environment_paths]
-      paths.any? && paths.all? { |path| File.exist?(path) }
+              options[:environment_paths]
+      paths.flatten!
+      if paths.any?
+        tst = true
+        paths.each do |path|
+          if !path || path.nil? || !File.exist?(File.expand_path(path))
```
jaymzh: call paths.compact to drop the nil's, then you can drop the first two parts of this conditional

jaymzh:
I haven't looked too deeply into the failing tests, but it looks like you're messing up option parsing somewhere in the help case because `No such file or directory --no-such-role-dir`

chazzly:
I finally got the errors resolved.  Any chance this can be considered for 5.0?

chazzly:
re-based against master.

chazzly:
Sorry for the multiple builds. I can't seem to get my rubocop set-up to use the same rules, so this was my only way to track them down.

I've rebase against master (7.0.1).  Any chance we can revisit this?

svmastersamurai:
Wow, I was browsing around wondering why foodcritic didn't have this configurability (if that's a word), like its comrade in arms `rubocop`, and here it is! 🍪
What would the chances be of this being revived and merged? It might help ditch some awful thing written in PHP I use on a daily basis for linting cookbooks in favor of a purely ruby solution 😉

tas50:
Closing this out since it'll need a significant rebase at this point. If it was rebased and the tests pass, I'd be down to get this included.

PR #294, opened 2015-01-12
Add metadata for missing chef 11 versions

jdmundrawala:


lamont-granquist:
:+1:  i imagine we need this for chef-12.0.x as well.

lamont-granquist:
paging @jaymzh

jdmundrawala:
@lamont-granquist i think chef 12 is going to require some work. It looked like switching to ruby 2 produced a lot of warnings when running the generate task. Will look into it more when I get a chance.

btm:
this looks sadly manual. :(

@acrmp I guess this is the right thing to do?

acrmp:
@btm What looks manual? If the generation can be made to work on Chef 12 then that would be the best way forward.

david-crowder:
Is there anything I can do to help with this issue? It seems like all of the comments specific to chef 11 have been addressed. Though 11.18.x seems to have come out since this PR was made. 
I have an interest in this because I use foodcritic and I have started to use the sensitive attribute in cookbooks, which sadly the current release of foodcritic does not support.

lamont-granquist:
mergifying

btm:
@acrmp I guess I was surprised we've got to run something and ship a release for this (and should be doing so regularly).

jdmundrawala:
@btm i think we should build the artifacts foodcritic needs as part of the chef client build

PR #299, opened 2015-02-06
Ensure metadata specifies maintainer and maintainer_email

nathenharvey:
Adds two new rules

* FC052 => Ensure maintainer is set in metadata
* FC053 => Ensure maintainer_email is set in metadata

See also [https://github.com/chef/chef-manage-issues/issues/20](https://github.com/chef/chef-manage-issues/issues/20)


odcinek:
* Rename to FC054/FC055
* Add a test?

lamont-granquist:
duplicates #248?

acrmp:
Hi Nathen,

Thanks for this. I rebased and backfilled tests.

We already have FC031 for the case where there is no metadata so I think we'll avoid these two new rules also matching on that case.

Thanks,

Andrew.

PR #300, opened 2015-02-18
Heredoc string interpolation erroneously triggers 002 on MRI 2.2

ryansouza:
Includes a failing test case and adds ruby 2.2.0 to travis runs where this fails


ryansouza:
We found foodcritic starting to complain on an internal cookbook when upgrading our travis runs to MRI 2.2. It seems to trigger on any usage of interpolation in heredocs, which should be fine to do.

I'm a little surprised the travis run failed on pre-2.2, I figured something about the AST had changed, maybe it was something else.

lamont-granquist:
The 1.9.3 travis error should go away if you rebase on master, the other error looks related to this change though.

PR #305, opened 2015-03-06
Update rules.rb

michnmi:
According to the man pages of wget: 
       Wget - The non-interactive network downloader.
This indeed is a downloader which should be constrained in a remote_file resource. 
However , 
According to the man pages of curl: 
curl is a tool to transfer data from or to a server, using one of the supported protocols
It nowhere explicitly is used as a downloader. 
It should be removed by definition from this rule. 
We could amend it by checking for curl  -o/--output <file>
This has been pointed out in : 
https://github.com/acrmp/foodcritic/issues/240
I would like to change this across food critic with a pull request. Please reply if you think that's acceptable

### DIFF
--- a/lib/foodcritic/rules.rb
+++ b/lib/foodcritic/rules.rb
@@ -618,7 +618,7 @@ rule 'FC041', 'Execute resource used to run curl or wget commands' do
   recipe do |ast|
     find_resources(ast, type: 'execute').select do |cmd|
       cmd_str = (resource_attribute(cmd, 'command') || resource_name(cmd)).to_s
-      (cmd_str.include?('curl ') || cmd_str.include?('wget '))
+      (cmd_str.include?('wget '))
     end
   end
 end


jaymzh:
The point isn't that you can never use wget/curl for other things, of course you can... the point is the most *common* use for both is for HTTP GETs of files, and thus there's a high likelyhood you're doing it wrong.

To me, if you're doing this it makes more sense to use `# ~FC041` on that line, or to just disable that rule in your config.

michnmi:
Hi Phil,
I beg to differ. The most common use for curl is not the GET of files. curl
is used to communicate with different protocols. Especially APIs. That is
its system usage.
I understand the ability to remove the rule. I also understand the ability
to change the rule.
However, FC41 suggests that using curl is only acceptable under remote_file
which implies that curl is used ,as you said, for the downloading of
files.  The rule is based on an assumption that is not correct based on the
very definition of cURL itself and its usage throughout the years by
SystemEngs and admins. It seems good practice to amend the rule to reflext
that by explicitly checking for the presence of the output file option.

jaymzh:
By default, with no options it grabs stuff:

curl www.google.com

Grepping a few codebases, I see tons of that. And people suck at using command-line options, so I imagine there's tons of `curl www.whatever.com/thing.txt > thing.txt` type things around.

That said... I'm not opposed to looking for the -o option...

lamont-granquist:
I'd also prefer to see http_request extended and/or have better DSL methods to use Chef::REST::Simple for doing API calls so that chef doesn't shell out to 'wget' or 'curl' to do that kind of work either.

odcinek:
As @jgoldschrafe pointed in https://github.com/acrmp/foodcritic/issues/240, checking for `-o` / `--output` and output redirection should be ok. That will fix most false positives, and the rest can be exempt.

odcinek:
@michnmi you want to rework this? Don't forget about test cases if you do.

michnmi:
@odcinek I saw you opened a new request on this. If you want to do it I'm  fine with that.  I was just waiting  on the debate to finish :)

PR #309, opened 2015-03-19
add warnings for use of recommends/suggests

lamont-granquist:
These are unimplemented metadata keywords and their use is currently a
sort of no-op.  However, this can cause bugs with different tooling:

https://github.com/berkshelf/berkshelf/issues/1404

It also is likely that whatever the user thinks the semantics should be
eventually that some subset of users will guess incorrectly and get
surprising behavior if we ever do implement these keywords.  The proper
approach to documenting optional cookbooks is currently via text in the
README.md or a comment in the metadata.rb.

The correct course of action in the future is that an RFC should be
submitted to https://github.com/chef/chef-rfc to decide what the suggests
and recommends semantics are, and to actually implement those semantics.
Until then, users should avoid the use of these keywords (essentially
they are 'reserved for future use' -- which means "do not use them now").

### DIFF
--- a/features/052_check_for_metadata_using_suggests_keyword.feature
+++ b/features/052_check_for_metadata_using_suggests_keyword.feature
@@ -1,0 +1,15 @@
+Feature: Check for metadata using suggests keyword
+
+  In order to ensure that recipe metadata is stable
+  As a developer
+  I want to identify metadata that is using unimplemented features whose definitions may change in the future
+
+  Scenario: Metadata with the suggests keyword
+    Given a cookbook with metadata that includes a suggests keyword
+     When I check the cookbook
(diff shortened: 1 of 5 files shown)


lamont-granquist:
@acrmp @jaymzh @fnichol @danielsdeleo could use some review.   i'd like to make this policy to avoid the use of these keywords until they actually become defined.

odcinek:
I would not mind having this as a single rule. Other than that lgtm.

lamont-granquist:
Yeah, my thoughts on two rules is that we might implement one of them before the other.  I think I looked at implementing `recommends` but didn't have any idea of what to do with `suggests` at one time.  And with two rules we could globally disable one or the other independently.

odcinek:
:+1:

PR #311, opened 2015-03-21
Add chef 12.0.0 through 12.1.1

odcinek:
Closes https://github.com/acrmp/foodcritic/issues/307


jaymzh:
sidenote - we need to start setting our version in foodcritic at FB.

:+1:

odcinek:
I was going to bump the default version but tests fail on anything `~> 11.18.0`, needs looking into

PR #312, opened 2015-03-21
Recognize force_default and force_override.

### LINKED ISSUE #181: FC019: force_default and force_override not allowed?
In a wrapper-cookbook, these two error out for me:

```ruby
node.force_override['base_system']['user']['default_password'] = password.join
node.force_default['ntp']['servers'] = %w[
  0.nl.pool.ntp.org
  1.nl.pool.ntp.org
  2.nl.pool.ntp.org
  3.nl.pool.ntp.org
  0.uk.pool.ntp.org
  1.uk.pool.ntp.org
]
```

not sure why they get caught in the FC019 rule. The error goes away if I swap both to `override` and/or `default`. Why is `force_*` not allowed in this case?

coderanger:
These were added in Chef 11.0.0.

This will fix #181.


odcinek:
Why were those not generated by `:generate_chef_metadata`? O.o

coderanger:
I had no idea that existed so I did the the normal patch-y way :-)

odcinek:
There must be something missing from `foodcritic/chef_dsl_metadata/Rakefile`, as it's not covering everything. Oh well.

JamesMGreene:
Any chance we can get a new release published?  This particular bug is driving me crazy... I've added `# ~FC019` ignores for now but I hate to do that.

lamont-granquist:
I accidentally reverted this so fixed it the right way in #364

PR #315, opened 2015-03-31
Default to recent chef version

odcinek:
In order to properly support 12-specific rules (and run their tests) I'm bumping the required chef version to `12.1.1`

This does *not* pull 12 as dependency. The only change users might be seeing is `--chef-version` being required (since `DEFAULT_CHEF_VERSION` changes from `11.16.4` to `12.1.1`).

`lucene.treetop` was being supplied with by 10.x dependency. In order to move forward I'm copying it over.


odcinek:
Failure on `1.9.3`:
```
Gem::InstallError: ohai requires Ruby version >= 2.0.0.
An error occurred while installing ohai (8.2.0), and Bundler cannot continue.
Make sure that `gem install ohai -v '8.2.0'` succeeds before bundling.
```

odcinek:
The search/treetop/lucene might be better handled by reusing parser from `chef-zero`, as stated here https://github.com/acrmp/foodcritic/issues/199.

lamont-granquist:
Yeah we've dropped support for 1.9.3 in Chef 12 / Ohai 8.x

Its EOL and no longer getting security backports:

https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/

I'd recommend dropping 1.9.3 from travis

odcinek:
We'll do a major release with `1.9.3` support dropped. @lamont-granquist @jaymzh - merge?

lamont-granquist:
:+1:

PR #319, opened 2015-04-10
Update README.md

jjasghar:
Having two foodcritic sites is really confusing. The foodcritc.io site seems to be more up-to-date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 Food Critic is a lint tool for Chef cookbooks. It requires Ruby 1.9.3+.
 
 # Homepage
-[http://acrmp.github.com/foodcritic/](http://acrmp.github.com/foodcritic/)
+[http://foodcritic.io/](http://foodcritic.io/)
 
 # Building
 


odcinek:
Since we'll be doing release soonish (https://github.com/acrmp/foodcritic/issues/318) it might be worth looking into both sites, and updating.

Why are there two in the first place?

jjasghar:
Completely agree it looks like the foodcritic.io is the "more up-to-date" one. I'm about to put in another PR to get that site up-to-date with the new URLs too.

jjasghar:
I added this also to help with the consolidation: https://github.com/acrmp/foodcritic-site/pull/7

odcinek:
@acrmp it would be nice to unify this, maybe keep one site? do you have any preference for how should we move forward?

acrmp:
@odcinek The GitHub pages site is the older site. It's deprecated so we'll want to update it to forward users to foodcritic.io.

loveleshsaxena:
Hi guys,
This is Lovelesh Saxena. I am working for SAP Labs. I want to develop few
food critics rules for my cookbooks. Please tell me, is it the right place
to ask some question.


Lovelesh.

PR #321, opened 2015-04-13
merge default options before check instead of during intialization

ranjib:
So that following can work:
```ruby
FoodCritic::Rake::LintTask.new do |t|
  t.options = {
    fail_tags: ['any'],
    cookbook_paths: ['site-cookbooks/', 'cookbooks/'],
    chef_version: Chef::VERSION
  }
end
```


jaymzh:
I'm fine with this. I'm going to change your title though - please don't ever use "default attributes" in a chef space unless you mean node attributes at a default precedent. :)

PR #322, opened 2015-04-16
Corrects new FC054 if run in current dir

chazzly:
When run from within the cookbook directory  "foodcritic .", FC054 was incorrectly trying to match the cookbook name to "." rather than the name of the current directory.


lamont-granquist:
reverted FC054 so closing this

PR #328, opened 2015-05-08
Add self-dependency warning

lamont-granquist:
Chef does not require cookbooks to depend on themselves in order to
depsolve.  In addition to this line being useless, it actually creates
problems with depsolving.  The self-dependency can create a circular
dependency where every version of a cookbook will satisfy the circular
dependency on itself which can create a huge mess.

Ideally we need to fix this elsewhere in the ecosystem as well, but
for now this will at least surface the issue to cookbook authors who
use foodcritic.

### DIFF
--- a/features/063_cookbook_incorrectly_depends_on_itself.feature
+++ b/features/063_cookbook_incorrectly_depends_on_itself.feature
@@ -1,0 +1,15 @@
+Feature: Check for metadata using suggests keyword
+
+  In order to avoid pathological depsolving issues
+  As a developer
+  I want to identify cookboks that depend on themselves and remove that (unnecessary) dependency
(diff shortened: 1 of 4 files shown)


lamont-granquist:
@acrmp @jaymzh this rule is at least 3 years overdue

thommay:
:+1:

tas50:
@lamont-granquist We should add this for the 5.1 release.  Update the rule number and lets get this merged.

tas50:
@lamont-granquist Travis is failing

lamont-granquist:
@tas50 still had a couple of references to "055" that needed to be "063" instead.  fixed.  squashed.

PR #342, opened 2015-07-09
Add output to built-in rake task

patcon:
What it says on the tin :)


patcon:
Helpful for when you're looking at CI output but all the lint tests pass

lamont-granquist:
Probably want to add a verbose flag like rubocop does at some point, but good enough for now.

PR #346, opened 2015-08-03
updated hash syntax to 1.9

askl56:


lamont-granquist:
This isn't critical and would be better handled by getting foodcritic sane in general

jaymzh:
In general the new ruby has syntax is horrible.

```
# hashes work
irb(main):003:0> foo = { :foo => 'bar' }
=> {:foo=>"bar"}

# new hashes, not so much
irb(main):004:0> foo = { :foo: 'bar' }
SyntaxError: (irb):4: syntax error, unexpected ':', expecting =>

# but certainly I can be more explicit:
irb(main):005:0> foo = { :"foo": 'bar' }
SyntaxError: (irb):5: syntax error, unexpected ':', expecting =>

# no, OK it likes strings, that's cool, what about this string?
irb(main):006:0> foo = { ":foo": 'bar' }
SyntaxError: (irb):6: syntax error, unexpected ':', expecting =>

# what the fuck? I'm not crazy, right?
irb(main):008:0> foo = { ":foo" => 'bar' }
=> {":foo"=>"bar"}

# OK, sweet, that works.
irb(main):010:0> foo = { \:foo: 'bar' }
SyntaxError: (irb):10: syntax error, unexpected $undefined, expecting '}'

# Damn. I give up.
```

PR #349, opened 2015-08-11
Removes what is no longer at opscode-cookbooks

philoserf:
These cookbooks, removed from the regression list, are no longer among those in the Chef Community Cookbooks organization.


philoserf:
Closing this as #348 and #350 should be part of the same commit.

PR #350, opened 2015-08-11
Removes expected results for missing cookbooks

philoserf:
The expected results for cookbooks no longer hosted at opscode-cookbooks removed.


philoserf:
Closing this as #348 and #350 should be part of the same commit.

Issue #344, opened 2015-07-13
Time to cut a release?


philoserf:
I believe time has come to cut a release.

tas50:
+1

spenserpothier:
+1

acrmp:
It's been a while. I'll look into this in the next couple of days.

acrmp:
I did look into this but there were some failures in the regression tests that will block cutting a new release.

philoserf:
Travis reports success: https://travis-ci.org/acrmp/foodcritic

I just ran the regression and it appears to me it fails because cookbooks in <https://github.com/acrmp/foodcritic/blob/master/spec/regression/cookbooks.txt> have moved or died.

Is the required effort then to update the regression cookbook list?

philoserf:
I am saving this result for comparison once I have regression results on master with the missing cookbook expected results removed.

FoodCritic @ tag v3.0.3 failed regression after removing missing cookbooks and missing cookbooks expected results

```
bundle exec rake regressions
Run options: --seed 27464

# Running tests:

F

Fabulous tests in 240.098212s, 0.0042 tests/s, 0.0042 assertions/s.

  1) Failure:
test_0001_should result in the expected matches against a pinned set of cookbooks(regression test) [/Users/pair/github/foodcritic/spec/regression/regression_spec.rb:14]:
--- expected
+++ actual
@@ -364,7 +364,6 @@
 FC034: Unused template variables: ./activemq/templates/default/wrapper.conf.erb:1
 FC034: Unused template variables: ./application_java/templates/default/context.xml.erb:1
 FC034: Unused template variables: ./application_python/templates/default/celeryconfig.py.erb:1
-FC034: Unused template variables: ./nginx/templates/default/nginx.init.erb:1
 FC034: Unused template variables: ./sbuild/templates/default/mk_chroot.sh.erb:1
 FC041: Execute resource used to run curl or wget commands: ./hadoop/recipes/default.rb:33
 FC042: Prefer include_recipe to require_recipe: ./postfix/recipes/aliases.rb:17


1 tests, 1 assertions, 1 failures, 0 errors, 0 skips
Coverage report generated for RSpec to /[snip]/github/foodcritic/coverage. 243 / 799 LOC (30.41%) covered.
rake aborted!
[...]

PR #358, opened 2015-09-10
Revert "FC054, check for mismatched cookbook names"

lamont-granquist:
This reverts commit df3c89075bb407d8e96d7aef44e16a33c89fce78.

Chef 12 requires the name attribute so that the directory name may be renamed, so FC054 gets the spirit of requiring the name attribute backwards.

See discussion in #344


lamont-granquist:
@danielsdeleo do you concur that FC054 should be pulled?

danielsdeleo:
I'm pretty sure I agree, but I'm a little fuzzy on how well we support things like having multiple same-named cookbooks in a chef repo and using knife to upload them, which could be a reason you would care. What was the situation in which you encountered this? Also, I looked at the commit you referenced, which says:

> In Chef 12 name metadata attribute needs to match cookbook dir name

which doesn't seem to be correct to me? I'm just all confused now :crying_cat_face:

lamont-granquist:
Right, its not correct.

And we may have areas in knife where we expect the cookbook dir to match the cookbook name, or where we trust the cookbook dir over the cookbook name, but those are bugs in knife and/or just issues with the tool where the tool should be responsible for whining about the problem if it really can't be changed.

As far as foodcritic goes, I don't think we want to be enforcing this and should be encouraging those bugs to get fixed in the tooling.

coderanger:
Two notes, we should make sure that with the rule removed that doesn't break people that have filtered out this rules and we should make sure we never add a new `FC054` for the same reason. Because of the latter it might be better to convert it to a stub rather than remove it fully?

danielsdeleo:
@coderanger's suggestion seems good, is that how we've handled removals in the past?

Anyway, I'm convinced this rule can be removed, whatever the correct process is.

coderanger:
A comment in the rules file would probably work too, I have no strong opinions on how to accomplish that piece of things :)

lamont-granquist:
comments are added.

coderanger:
:+1: :shipit:

lamont-granquist:
Only prior art I can find is when FC001 was yanked and then restored, and the commit that yanked it just deleted it like this one does...

lamont-granquist:
If this causes issues we can look at stubbing or ignoring overrides for rules that don't exist...

jaymzh:
We added this because it bites us in the ass all the time. Someone does

```
cp -pr foo/ bar/
# update bar/recipes/default.rb
# commit
```

And then we spend 4 fucking hours trying to figure out why on earth foo is doing such weird things.

jaymzh:
I'm so goddamn tempted to revert this commit. You sent the PR and merged it in less than an hour. However, I'm going to wait until tomorrow morning so I don't do what you did.

Rules are configurable and people don't have to use them all... but this is a totally valid rule that can catch a horrible breakage.

chazzly:
I'm not Collaborator status here, but I agree with @jaymzh.  One of the beauties of this tool (and arguably it's primary purpose) is how is helps us avoid problems and mistackes, which includes not confusing yourself and your teammates.  Yes, we, as community should encourage the tool to establish a consistent source of truth (metadata or dir name) and rely only on that source of truth, but that doesn't change the fact that having a cookbook named bar in a directory named baz would confuse the hell out of most people, and just seems like you're asking for trouble.

The answer may be in @lamont-granquist's comment at the beginning of this thread:
```"FC054 gets the spirit of requiring the name attribute backwards."```

I thinkg there are two different issues trying to be addressed:
1. (correctness) Chef 12 requires the name attribute, so it must be properly defined and maintained.
2. (style) You will avoid a load of confusion and potential problems, if your directory names match your cookbook names.

The first sounds like FC045 needs to be worded more strongly and labeled as correctness rather than annoyances.
This (FC054) addresses the second and is valuable.

My $0.12  - inflation of $0.02 from 1971 to 2015   :-)

**Comment by jaymzh** [comment]
Well put. To put my less-angry more-explain-y post from #344 here:

```
Chef lets you do LOTS of things that aren't preferred that foodcritic warns you about. Like, for example putting an if around a resource. Sometimes that IS what you want, but often it's not, so we warn on it. Your argument that "it's supported" is invalid - of course it's supported... but it's LIKELY to be an error. Which is where linters come in.
```

This is the case of the thing where what's in metadata.rb is what's used and it's a **really good idea** to have the directory be named the same thing or your shit will probably break. And if you really want them to be different, to quote @lamont-granquist  from another bug yesterday, "then you know enough to disable the rule."

danielsdeleo:
Correct me if I'm wrong, but it sounds like the actual thing you want to avoid is having two cookbooks with conflicting metadata?

jaymzh:
What does that even mean? How would you know they are two different cookbooks? We specifically lint cookbooks individually (plus their depends) to catch missing dependencies, for example.

Though as a theoretical check, sure that would be nice, but no, it's not what I "really want", it's a neat additional thing.

What I really want is to NEVER upload a cookbook that looks like one thing but really is another thing. Lint is about best practice. Best practice is to name the cookbook directory the name of the cookbook.

lamont-granquist:
The problem that I see is that this is a very context dependent rule.  It already bit @coderanger and he disabled all those checks.  All you have to do is download the cookbook to a different directory and foodcritic will fail the rule, making the rule not very portable.  I suspect that we'll immediately have to figure out how to disable the rule on supermarket.

I see your point, but I think it'll also cause lots of issues in fairly normal use cases.  I'd kind of like to know what @coderanger thinks since he already hit this one in actual practice.

jaymzh:
When you download cookbooks ... it should go into a directory... as a directory. The cookbook is the entire cookbook directory. So if you download a CB to `/tmp/test` it should come down as `/tmp/test/the_awesome_cookbook`. If it doesn't, I'd say the tools are broken. You're welcome to disable rules you don't like. I disable the god awful 1.9 hash syntax rubocop rules and we disable about 10 or so FC rules... but this is definitely best practice for production.

coderanger:
The reason I have it disable is all my cookbooks live in temp folders so matching the name isn't an option. This isn't really a "best practice" in the way most of the others are, it's a workflow-specific error check.

coderanger:
It is also unlike all other rules as when I run `foodcritic .` I would not expect it look at anything outside of `.` but this does.

lamont-granquist:
I'd be more inclined to have this rule in the codebase but disabled by default, then it could be enabled for specific workflows.

lamont-granquist:
I'm not sure foodcritic has the concept of disabled by default, though, so that would probably be an additional feature patch...

jaymzh:
having all our cookbooks in temp folders means they're not in a repo, which seems inandofitself poor practice. Thou shalt never edit things that are not in revision control.

coderanger:
@jaymzh In my case the actual cookbook is compiled output, not the thing in version control :) To be clear, my case is very much not the norm, but I still disagree with calling this a best practice. If you name your SCM folders something like `~/chef-whatever` that would be totally in line with all best practices even when the cookbook is named `whatever`.

lamont-granquist:
A lot of tooling also uses the_awesome_cookbook-1.2.3 and i'm pretty sure the tarballs uploaded to the community site to share have that as the directory.

lamont-granquist:
The mandatory cookbook name stuff also came from the Berkshelf world and @sethvargo might be able to comment on that since he's also still a maintainer here as well...

jaymzh:
@lamont-granquist  just like with any other software, when you cut a release you can make a directory with a given name. chef-client's repo doesn't have a directory called 'chef-12.4.1' ... it's just chef. When you export it from SCM to make a release you call it chef-12.4.1.

coderanger:
Yes but running foodcritic on a downloaded zip file from the Supermarket should work. Basically folder name == cookbook name doesn't apply in all cases and there isn't a compelling reason to tell people they should _always_ match. I could see maybe saying that `folder_name.include?(cookbook_name)` is a more general rule, but I would still personally leave it disabled.

jaymzh:
There isn't a compelling reason to tell people you should NEVER wrap an `if` around a resource either... SOMETIMES it's the right thing to do. There isn't a compelling reason to tell people to NEVER use a definition, sometimes it's the right thing to do. Your statement is totally valid, but completely orthogonal to whether this should be a rule.

Hmmm - maybe `startswith?`

coderanger:
There are downsides to not following those rules that are fairly universal whereas the mentioned folder names have zero downside outside of the specific workflow you mentioned.

`starts_with?` would block folder names like `chef-whatever` which IMO are fine. I've pretty often had `~/src/myapp` for the application and `~/src/chef-myapp` for the cookbook to deploy it.

jaymzh:
I disagree... there's totally valid needs for `if` around a resource... 

I suppose include? is probably good enough...

lamont-granquist:
I use `if` around resources to avoid putting the resource on the resource collection because its just not an appropriate resource for the converging server.

The thing is that foodcritic works on that rule no matter what my workflow is.  When I override it, it is also fully encapsulated in the state of the cookbook and the override is also invariant and everyone who asks the same question gets the same answer.  This rule is pretty odd because like @coderanger pointed out it looks outside of '.' and depending on the different workflows you use, you can get a different answer.

This might be a good topic for discussion at the next community meeting.

coderanger:
I think with `include?` this would be generic enough to be enabled by default. I still would put it on a different tier than most of the other checks. I guess this is why Rubocop has `error` vs. `warning` vs. `convention` levels, but this is not the time or place to talk about adding that kind of thing :)

lamont-granquist:
`starts_with?` is also vulnerable to `name "foo" `in a "foobar" directory matching and not triggering this rule when it should.   we could use foo-\d+-\d+-\d+ since cookbook versions are enforced to be 3 integers (although i really think we should change that at some point).

lamont-granquist:
`include?` is also vulnerable to even more false negatives...

coderanger:
@lamont-granquist We need to match `name`, `chef-name`, `name-chef`, and `name-1.2.3` at a minimum as those are all relatively common. I'll take false negatives over false positives given the nature of the check.

jaymzh:
I like the idea of explicitly allowing common permutations...

lamont-granquist:
I'm still thinking this kinda bikeshed is why we have the community meetings...

coderanger:
Sure, just trying to come up with a minimum check that will not throw errors for most people while still covering the specific case of "copy the cookbooks directory and forget to update the name" since that seems like the most important thing this will flag. `include?` seems like it fits that bill and would unblock a release.

lamont-granquist:
Well, we can release it like it is and add FC054 back for 5.1.0 a week later or so (we could revert the revert for 5.0.0 and then revert the revert of the revert and release a 6.0.0 if we decide to a week later, but that seems like it would be more confusing to people).

jaymzh:
@lamont-granquist  - too many pronouns - is "this" the above proposal, or with the revert, or as the original code was? I'd like to release with [some form of] this rule in there since we use it.

thommay:
right, let's back out this reversion so that we ship 5.0.0 with FC054 included as-is.

lamont-granquist:
yeah, after thinking about it over the weekend, what i'd really like to do in a perfect world is ship FC054 disabled, but FC doesn't have good support for optional rules and needs something like #290 merged at a minimum.  so given the choice between on-but-disable'able and not supported at all, i think it should be on, so so we revert the revert.  will work on that later today after the caffeine kicks in a bit more...

adamhjk:
This is a good discussion. Clearly, some folks like and need this rule in order to prevent what are common breakages in their environment, while for others, this would break their existing (and supported explicitly by Chef 12) workflows. Here is how w'ere going to resolve it:

* We are not going to ship this FC054
* We are going to extend foodcritic to handle opt-in rules, rather than exclusively opt-out
* Then we are going to merge FC054 as a default opt-in

Our reasoning here is two-fold:

1. You can extend foodcritic with custom rules today relatively easily
1. This rule has not shipped in a release, therefore it will do no harm in the broader community
1. Clearly we need to improve what rules can ship stock in foodcritic

Deciderated.

Love,
Adam

PR #362, opened 2015-09-10
FC009 chef metadata update

lamont-granquist:
- adds latest versions to Appraisals
- expunges Appraisals of version that no longer build
- updates metadata of older versions for ruby 2.2.x
- adds metadata for all the new versions
- adds a README so next time the process should be easier to remember


lamont-granquist:
For reference this was a subtask off of #344

PR #375, opened 2015-09-16
fixes for cookbooks.txt and expected-output.txt

lamont-granquist:
remove more cookbooks not in chef-cookbooks
fix up more stuff in expected-output.txt


tas50:
:+1:

PR #376, opened 2015-09-16
fix FC054

lamont-granquist:
foodcritic '.' was refusing to match so added File.expand_path plus a
test.


lamont-granquist:
This turned up in one of the regression tests, turned it into a cuke feature as well

tas50:
:+1:

PR #385, opened 2015-09-28
Support old-style notifications with ruby 1.9 hash syntax

akihiro17:
`foodcritic` raises `ArgumentError` if we run `foodcritic` on recipes that contains old-style notifications with ruby 1.9 Hash syntax.

```ruby:default.rb
service 'ssh' do
  action :nothing
end

file '/var/hoge' do
  action :create
  notifies :restart, resources(service: 'ssh')
end
```

```
> bundle exec foodcritc default.rb
=> foodcritic/chef.rb:69:in `resource_check?': Arguments cannot be nil or empty. (ArgumentError)
```

This commit solves the above issue

### DIFF
--- a/lib/foodcritic/xml.rb
+++ b/lib/foodcritic/xml.rb
@@ -23,7 +23,20 @@ module FoodCritic
       child.each do |c|
         n = xml_create_node(doc, c)
         c.drop(1).each do |a|
-          xml_node.add_child(build_xml(a, doc, n))
+          if a.first == :@label
+            # if the ruby 1.9 hash syntax is used,
+            # the ast like below is generated.
+            # ast:
+            #  [:assoc_new,
+            #   [:@label, "service:", [6, 39]]]
+
+            # create a label node and add it to the accos_new children nodes
+            label_node = xml_create_node(doc, a)
+            n.add_child(build_xml(a, doc, label_node))
+            xml_node.add_child(n)
+          else
+            xml_node.add_child(build_xml(a, doc, n))
+          end
         end
       end
     end
--- a/lib/foodcritic/notifications.rb
+++ b/lib/foodcritic/notifications.rb
@@ -94,6 +94,9 @@ module FoodCritic
     def old_style_notification(notify)
       resources = resource_hash_references(notify)
       resource_type = resources.xpath("symbol[1]/ident/@value").to_s.to_sym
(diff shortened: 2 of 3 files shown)


lib/foodcritic/xml.rb
```diff
@@ -23,7 +23,12 @@ def xml_hash_node(doc, xml_node, child)
       child.each do |c|
         n = xml_create_node(doc, c)
         c.drop(1).each do |a|
-          xml_node.add_child(build_xml(a, doc, n))
+          if a.first == :@label
+            label_node = xml_create_node(doc, a)
+            xml_node.add_child(build_xml(a, doc, label_node))
+          else
+            xml_node.add_child(build_xml(a, doc, n))
+          end
```
lamont-granquist: can you explain what this is doing?  could probably use both a comment here, and a somewhat less terse explanation in the PR.  seems like this might be intrusive to other rules as well since it hits how we construct the XML representation of the AST.

tas50:
@akihiro17 I ran into this today for the first time. Any chance you can get back to lamont and also rebase this. I'd love to get Foodcritic released with a fix for this bug.

akihiro17:
This commit generates following xmls.

hash rockets

recipe
```ruby
notifies(:restart, resources(:service => "nscd"))
```

ast
```ruby
[:assoc_new,
  [:symbol_literal,
    [:symbol,
      [:@ident, "service", [6, 40]]
    ]
   ],
   [:string_literal, [:string_add, [:string_content], [:@tstring_content, "nscd", [6, 52]]]]
]
```

xml
```xml
<assoc_new value="string_literal">
  <symbol value="symbol">
    <ident value="service">
      <pos line="6" column="40"/>
    </ident>
  </symbol>
</assoc_new>
```

ruby 1.9 hash syntax

recipe
```ruby
notifies :restart, resources(service: "nscd")
```

ast
```ruby
[:assoc_new,
  [:@label, "service:", [6, 39]],
  [:string_literal, [:string_add, [:string_content], [:@tstring_content, "nscd", [6, 49]]]]
]
```

xml
```xml
<assoc_new value="string_literal">
 <label value="service:">
  <pos line="6" column="39"/>
 </label>
</assoc_new>
```

akihiro17:
@lamont-granquist @tas50 

 I have fixed the code and added some comments.

PR #388, opened 2015-09-30
Remove windows_package attributes

offslinker:
Windows cookbook provides additional return_codes attribute to windows_package resource. I haven't found easy way to skip check when windows cookbook is in metadata.


lamont-granquist:
Needs to be fixed upstream, correctly, by chef/chef#3037

PR #390, opened 2015-10-09
(On behalf of Lexmark) Fix merging default options

justinrstout:
https://github.com/acrmp/foodcritic/commit/1d280d72f7ac37b3b3f8290f06a08a6d86f2d276 inadvertently reversed the order that default options were merged, rendering the user of the Rake task unable to override the defaults.


lamont-granquist:
closed via #382

PR #393, opened 2015-10-14
Add windows_package.success_codes attributes

Annih:
This PR add `success_codes`  to `windows_package` recognized attributes list.

### DIFF
--- a/chef_dsl_metadata/chef_11.12.0.json
+++ b/chef_dsl_metadata/chef_11.12.0.json
@@ -10481,6 +10481,7 @@
       "source_line=",
       "state",
       "subscribes",
+      "success_codes",
       "supports",
       "taint",
       "tainted?",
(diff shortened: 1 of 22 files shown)


lamont-granquist:
Can't accept this.

These files are autogenerated via extracting properties off of resources in the core chef DSL.  If you manually patch them they'll get destroyed when these files are regenerated, and you'll lose it on every upgrade.

This property exists only on the windows_package resource which is in the cookbook and not in core chef.  The proper solution to this problem is to finally port all of the functionality of windows_package  in the cookbook into core chef, and to nuke the resource from the cookbook.  

Until then you'll need to continue to tag it manually in the cookbooks you're writing.

tas50:
Just a small side note: Adding the resources from the Windows cookbook into core is on the roadmap so the need for this also goes away in the next few quarters.

lamont-granquist:
s/next few quarters/next quarter/ according to btm on the linked to ticket.

Annih:
Nice, thanks for the explaination :)

PR #402, opened 2015-11-03
add checks for correct use of use_inline_resources

lamont-granquist:
- recommend use_inline_resource in all 'LWRPBase' providers
- catch the use of use_inline_resources when users are also declaring
  `def action_whatever` instead of `action :whatever` which defeats
  use_inline_resources

### DIFF
--- a/features/057_check_for_library_provider_without_use_inline_resources.feature
+++ b/features/057_check_for_library_provider_without_use_inline_resources.feature
@@ -1,0 +1,20 @@
+Feature: Check for library providers that do not declare use_inline_resources
+
+  In order to ensure that notifications happen correctly
+  As a cookbook provider author
+  I want to always use_inline_resources
+
+  Scenario: Library provider with use_inline_resources
+    Given a cookbook that contains a library provider with use_inline_resources
+     When I check the cookbook
+     Then the library provider without use_inline_resources warning 057 should not be displayed against the libraries file
+
+  Scenario: Library provider without use_inline_resources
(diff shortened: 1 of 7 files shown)


coderanger:
:+1: as it is restricted to `LWRPBase`-based things.

lamont-granquist:
Yeah, I'm certainly intending to only target the bodies of classes that inherit from Providers::LWRPBase, we'll have to see how buggy my Ruby AST XPath parsing foo is -- but it does have a test that it avoids the simple case of a library file with only a resource in it, which does not declare use_inline_resources and ignores that.  I'm not sure I could actually do any 'better' than that, since we're doing static analysis.

There might be edge cases where an abstract base class inherits from LWRPBase but doesn't declare use_inline_resources, even though all the subclasses do, but I think the solution there is to either move use_inline_resources up to the baseclass or else just comment out the FC rule.

lamont-granquist:
@jaymzh  @tas50 etc got any thoughts to add about this one?

features/058_check_for_lwrp_provider_bad_action_thing_methods.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Check for LWRP providers that do not declare use_inline_resources
```
tas50: Looks like this description didn't get filled out

philoserf:
I approve AND I prefer on rule per Pull Request.

:+1:

lamont-granquist:
rebased, squashed, fixed the merge conflicts over the numbering...

PR #404, opened 2015-11-06
Require Oracle as a RHEL equiv

tas50:
It's in the same platform family so it should really be part of this rule


tas50:
I need to update the tests for this since a few more cookbooks would trigger this alert now.

lamont-granquist:
:+1: when you fix the tests

lamont-granquist:
:+1:

PR #405, opened 2015-11-06
New Rule 61 - valid cookbook version

lamont-granquist:
closes #351

### DIFF
--- a/lib/foodcritic/rules.rb
+++ b/lib/foodcritic/rules.rb
@@ -828,7 +828,7 @@ rule 'FC060', 'LWRP provider declares use_inline_resources and declares #action_
 end
 
 rule 'FC061', 'Valid cookbook versions are of the form x.y or x.y.z' do
-  tags %w{metadata}
+  tags %w{metadata correctness}
   metadata do |ast, filename|
     # matches a version method with a string literal with no interpolation
     ver = ast.xpath(%Q(//command[ident/@value='version']//string_literal[not(.//string_embexpr)]//tstring_content/@value))
--- a/lib/foodcritic/rules.rb
+++ b/lib/foodcritic/rules.rb
@@ -826,3 +826,28 @@ rule 'FC060', 'LWRP provider declares use_inline_resources and declares #action_
     end
   end
 end
+
+rule 'FC061', 'Valid cookbook version should be defined' do
+  tags %w{metadata}
+  metadata do |ast, filename|
+    ver = ast.xpath(%Q(//command[ident/@value='version']/descendant::tstring_content/@value))
+    if ver.nil? || ver == ''
+      [file_match(filename)]
+    else
+      pass = true
+      parts = ver.to_s.gsub(/\s|'|"/, '').split('.')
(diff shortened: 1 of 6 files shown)


lib/foodcritic/rules.rb
```diff
@@ -826,3 +826,28 @@ def invalid_name(ast)
...
+rule 'FC061', 'Valid cookbook version should be defined' do
+  tags %w{metadata}
+  metadata do |ast, filename|
+    ver = ast.xpath(%Q(//command[ident/@value='version']/descendant::tstring_content/@value))
+    if ver.nil? || ver == ''
+      [file_match(filename)]
+    else
+      pass = true
+      parts = ver.to_s.gsub(/\s|'|"/, '').split('.')
+      if parts.length > 1
```
coderanger: Do we want to check `> 3` too?

lamont-granquist: Yeah, this is just a rebase of the existing code, I want to add tests and fix that.  CBVs actually have to be exactly 3 dotted decimals and this code can most likely be simplified considerably.

coderanger: Probably okay to accept one or two parts and let it extend them, or at least that is pretty common and doesn't seem like a bad idea.

lamont-granquist: Okay I didn't realize 'x.y' was acceptable:

```
ERROR: Chef::Exceptions::InvalidCookbookVersion: '1' does not match 'x.y.z' or 'x.y'
```

We should just do that.

lamont-granquist:
@coderanger @tas50 @acrmp hows this look now?

coderanger:
I would make throw in one test using double quotes because I'm paranoid, but :+1: :shipit:

acrmp:
A `metadata.rb` with this would also match incorrectly - don't know how common it would be in practice:

```ruby
expr = '1.0.0'
version expr
```

lamont-granquist:
Uhm, yeah, that's actually terrible and blocker for this.

We need to only select string or double string literals here and check them.  Peeps with CI/CD style cookbook delivery systems may be calling out to random ruby libraries to set their versions.

lamont-granquist:
Also we should probably not file_match but match with a line number...

lamont-granquist:
@coderanger @acrmp @tas50 think this one is ready.

lib/foodcritic/rules.rb
```diff
@@ -826,3 +826,21 @@ def invalid_name(ast)
     end
   end
 end
+
+rule 'FC061', 'Valid cookbook versions are of the form x.y or x.y.z' do
+  tags %w{metadata}
```
tas50: We should also include the tag correctness here

acrmp:
Thanks. I think FC061 would still match this case:

```ruby
version generated('like this')
```

In some rules I believe we count ancestors to avoid this kind of issue.

We could also return the matching node so that we get the correct line number rather than using `file_match`.

lamont-granquist:
working on a fix...

PR #406, opened 2015-11-10
bump version to 5.1.0

lamont-granquist:
- Update version to v5.1.0 for release
- Switch CHANGELOG.md to github_changelog_generator and add 'rake changelog' command to update it.


lamont-granquist:
@tas50 @acrmp look okay?

github-changelog-generator loses a little bit of detail, particularly on some fixes that were merged without PRs, but even in those cases the full diff is always available.  lets just switch and start annotating PRs better going forwards.

we can also fix the PR history retroactive to skip or update titles and the changelog will get updated.

tas50:
I think we shouldn't use the auto generated changelog.  We lose the historical changes and any explanation of the significance of the changes in versions.  At that point we might as well just delete the changelog and let people comb through Git logs to find out what changed.

lamont-granquist:
Yeah, but manual entry is a PITA and is a human typey-typey barrier to getting the job done.  I really don't like all the copypasta work updating this changelog.  Code >> Humans, always.

tas50:
The person on the other end consuming that is a human so a human should write it.  Otherwise it's just a git log and those are not helpful to humans.

lamont-granquist:
Realistically we don't, though, we just manually copy and paste the titles of the PRs into entries.  And, we can fix the titles of the PRs.  Really its about the same problem, but with less humans cutting and pasting.

tas50:
If you read the current changelog it's much more consumable for humans than a pure git log.  It beaks out the new rules and rule changes nicely. Not hard to write up.  You just comb through the git log and summarize in human readable form and note anything that's going to break / impact people.  We really should be doing this by hand.

danielsdeleo:
Just to be 100% clear, it's using the pull request titles and not just dumping the git log. We can go back and re-title any PRs that are unclear or add tags to issues to get them categorized. So long as we do that, I think it's fine and a lot less tedious.

randomcamel:
It's more human-friendly than the raw git log: take the [chef-zero changelog](https://github.com/chef/chef-zero/blob/master/CHANGELOG.md) as an example. It's pretty similar to what we do by hand, e.g. in [chef/chef](https://github.com/chef/chef/blob/master/CHANGELOG.md).

We can take the current manually-generated changelog and just move it aside to a different file, so any history there will be easily available. In the repos where we've used the generator, it's been a big jump in changelog clarity and completeness.

stevendanna:
TL;DR: The goal is to get consensus around a single process for consistently putting together releases. :+1: with a caveat: I'd prefer to not edit change log entries for past releases if possible. 

It is hard for me look at the *Unreleased* section of the new change log and not feel like we've lost something. The human-made change log has consistent wording for similar features, has better feature descriptions, and can be consumed without clicking many PR links. Similarly, the chef-zero change @randomcamel mentions is more nicely formatted than a raw git log, but isn't much more clear to a normal user reading it. Compare those to the change log for a substantially more complex project: https://github.com/systemd/systemd/blob/master/NEWS

This was going to be true of any automated tool. To get a human-readable change log that provides more value than a nicely formatted git log requires human effort at some point in the process, either at the time of writing commit/PR messages or at the time of constructing the change log, or both.

That said, right now a major stumbling block we have is the manual work that goes into releasing various projects and the idiosyncratic nature of each process.  Our overall goal is to arrive at a common release process.  As the author of a competing internal tool that has seen almost no adoption, I can confidently say that getting everyone using the tool/process is more important than it having the output we want *right now*.  We can improve the output over time. We can do manual edits to the generated changelog if we feel it is too unclear.  We can replace it with a different tool that slots into the same release process. However, all of these improvements are only effective once we are consistently using a similar process across a number of our projects.

Yes, the automated output is worse right now.  But the goal is to get better about releases across a broad set of our projects and this helps in that direction.

jkeiser:
The changelog entries are human edited (PR titles) and can be changed when we want. Fixing them actually makes both the PRs and the changelog better.

You can tell github_changelog_generator to start generating at a specific version and include an old changelog if you so desire. I suggest only doing this for projects with a *ton* of history.

acrmp:
I'm ok with the generated changelog assuming we retain the distinction between features and bugfixes through labels (which I believe the tool supports).

We could also move the existing changelog out of the way and have the generator append it:
https://github.com/skywinder/github-changelog-generator#migrating-from-a-manual-changelog

jkeiser:
Yep, it definitely supports feature / bugfix distinction.

tas50:
We should probably close this out until the various issues with the changelog generation are resolved.  I've come around on the overall idea of generating the changelogs, but we need to get a solid release before we can do that.

lamont-granquist:
Yep, that bug is pretty much a hard blocker i think

PR #408, opened 2015-12-02
fix for edge condition with 061

lamont-granquist:
differentiates between `version function('arg')` and `version '1.2.3'`
by being strict about the nodes that we traverse.

i'm not sure if there's still a way to attack this via adding siblings,
but its an improvement...


tas50:
+1

lamont-granquist:
This is clearly an improvement over master, so merging it.

@acrmp can you think of any other attacks against it?  Ways to introduce siblings in the AST?

PR #409, opened 2015-12-03
Require the latest rubocop

tas50:
If someone is doing dev work we should ensure they have a modern release

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,5 +12,5 @@ end
 group :development do
   gem 'chef', '~> 12.1'
   gem 'ronn', '~> 0.7'
-  gem 'rubocop', '~> 0.20', require: false
+  gem 'rubocop', '~> 0.35', require: false
 end


lamont-granquist:
currently i think this is pointless and we don't rubocop anything, and i'd vastly prefer to use finstyle over rubocop...

lamont-granquist:
also '~> 0.20' still matches '0.38.0' or whatever

my Gemfile.lock has this:

```
% egrep rubocop Gemfile.lock
    rubocop (0.35.1)
  rubocop (~> 0.20)
```

tas50:
we should get something in there and start enforcing via travis.  At the moment this does nothing, but bumping to to .35 ensured you had a recent release.  Let me get an actual ruleset and get this in travis first.

lamont-granquist:
doesn't really 'ensure', though, people still need to 'bundle update' correctly or their lockfiles will be out of date, a fresh bundle install will pull the latest.

PR #410, opened 2015-12-03
use_inline_resources checks apply to Chef 11+

tas50:


lamont-granquist:
doesn't this only run on 11+ these days?

anyway :+1:

Issue #365, opened 2015-09-10
FC048 regression in master


tas50:
I believe we broke the detection of system in FC048 with this commit:

lamont-granquist:
The behavior is master is preferable to the bug in #252 

As long as we're catching the use of system in not_if/only_if and directly in the recipes then this isn't a show-stopper for releasing.

tas50:
Works for me, but lets leave that in the expected output file so we note it as a regression.  It should get fixed sometime.

lamont-granquist:
okay sounds good

lamont-granquist:
(and i just validated it catches the not_if/only_if and recipe DSL cases still)

Issue #252, opened 2014-07-21
FC048 incorrectly matches User resource system attribute


baylisscg:
I've a User resource that looks like this

```
  user new_resource.name do
    home new_resource.home
    comment new_resource.comment
    system new_resource.is_system %w{linux}.include?(node["os"])
    action :create
  end
```

FC048 is matching on the `system ...` line incorrectly ( it's an attribute not a shell out).

PR #413, opened 2015-12-11
FC048: Warn within a provider block, refs #365

acrmp:
The fix for #252 (aa57cb69) didn't handle multiple nodes.


lamont-granquist:
:+1: @tas50

PR #414, opened 2015-12-17
Code climate support

someara:


tas50:
I'm going to close this one out at this point. If there's a way we can better support code climate within the single binary and without introducing a large amount of code change I'm all for it. This doesn't seem like the right path forward though.

PR #417, opened 2016-01-08
Update metadata

tas50:
Also update the gemfile locks for 2 chef versions that were using a yanked chef-config gem


lamont-granquist:
:+1:

PR #418, opened 2016-01-08
Release 6.0.0

tas50:


tas50:
@acrmp @lamont-granquist Thoughts on getting this pushed out the door?

coderanger:
What makes this incompatible with 5.0?

tas50:
Tags have been changed, which could/will impact users

lamont-granquist:
leo

PR #421, opened 2016-01-26
Check for URLs that are helpful to the Supermarket

nathenharvey:
This PR introduces two new rules:

* FC064 - Ensure issues_url is set in metadata
* FC065 - Ensure source_url is set in metadata

In order to be able to share my cookbook on the Supermarket
As a developer
I want to be notified when my cookbook metadata does not specify the source_url or issues_url

I realize that not everyone who uses Foodcritic does so with the expectation that their cookbook will be shared on the Supermarket.  As such, I'm proposing a new tag `supermarket` be added.  Currently, I've tagged these two rules with both `metadata` and  `supermarket`.

Implementing these two rules in Foodcritic will make the implementation of https://github.com/chef-cookbooks/cookbook-quality-metrics/ easier.

### DIFF
--- a/features/064_check_for_no_issues_url.feature
+++ b/features/064_check_for_no_issues_url.feature
@@ -1,0 +1,25 @@
+Feature: Check for no issues_url
+
+  In order to be able to upload my cookbook
+  As a developer
+  I want to be notified when my cookbook metadata does not specify the issues_url
+
+  Scenario: Metadata without a issues_url
+    Given a cookbook with metadata that does not include a issues_url keyword
+     When I check the cookbook
+     Then the metadata missing issues_url warning 064 should be shown against the metadata file
+
+  Scenario: Metadata with a issues_url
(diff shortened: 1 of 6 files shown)


coderanger:
These need to be disabled for versions of Chef which don't support the options.

nathenharvey:
thanks, @coderanger.  I've updated them to only apply to chef 12.0.0 or higher.

robbkidd:
I propose the proposed `supermarket` tag be maybe `collaboration`. Thinking that things like `issues_url` facilitate collaboration no matter whether the cookbook is destined for public/private Supermarket or not.

tas50:
:+1: Much needed.

nathenharvey:
Anyone have thoughts on @robbkidd's suggestion that the tag be `collaboration` instead of `supermarket`?  While my interest lies primarily with the Supermarket for this PR, I do think `collaboration` better captures the nature of these rules.

PR #423, opened 2016-02-17
Fix FC058 false positives

jaym:
There were multiple issues here. First, the test did not
correctly catch this because the wrong match specification
was being created due to line number not being specified.
Instead of defaulting to nil, it defaulted to 1, making
2 incorrect things look correct together.

Second, FC058 needed to check the length of the array
was greater than so, as `[]` defaults to truthy.

cc @lamont-granquist @tas50 @sersut

Should fix https://github.com/acrmp/foodcritic/issues/419

### DIFF
--- a/features/058_check_for_library_provider_bad_action_methods.feature
+++ b/features/058_check_for_library_provider_bad_action_methods.feature
@@ -12,14 +12,14 @@ Feature: Check for library providers that declare use_inline_resources and decla
   Scenario: Library provider without use_inline_resources and (okay) action_create
     Given a cookbook that contains a library provider without use_inline_resources and uses def action_create
      When I check the cookbook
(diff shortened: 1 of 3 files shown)


lamont-granquist:
I suspect that line number defaulting to 1 was intentional and it should be fixed somewhere else up the stack to use 'nil' more explicitly...

jaym:
I thought that as well, however it checks for nil in part of it

jaym:
See https://github.com/jaym/foodcritic/blob/jdm/058-false-positives/features/support/command_helpers.rb#L134

jaym:
I guess we can push it up the stack based on the order of that merge

jaym:
I actually disagree with that default being there, this is an example of where badness happened because of it

lamont-granquist:
:+1:  LGTM @tas50

nathwill:
thanks @jaym! can confirm this solved it for me :+1:

PR #425, opened 2016-02-22
Fix FC033 false positives

tas50:
To be honest I'm not entirely sure what was going on here before, but it
certainly seemed complex for what we're trying to do.  All we need to do
is see if anything in the templates match the sources defined template
file. We have a method for grabbing all the templates and it does the
right thing out of the box returning templates that are in the root of
the templates dir (vs. in default or what not).  I tested this against
the working cookbook and then put in a bogus template source and it
triggered there as expected.

### DIFF
--- a/features/033_check_for_missing_template.feature
+++ b/features/033_check_for_missing_template.feature
@@ -1,4 +1,4 @@
-Feature: Check for missing template
+Feature: Check for missing template source file(s)
 
   In order to ensure the Chef run is successful
   As a developer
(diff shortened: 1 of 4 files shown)


tas50:
I'm baffled by this one at this point entirely because I have no idea how cucumber works.  @lamont-granquist or @acrmp any input would be greatly appreciated

lamont-granquist:
Yeah I'm not quite sure what you're not grok'ing about cucumber, but its clear you're not grok'ing it.

Those different template resources should all go into isolated 'Given' examples.

```
  Scenario Outline: Template types
    Given a cookbook recipe that <template_type>
     When I check the cookbook
     Then the missing template source file(s) warning 033 <warning>
  Examples:
    | template_type                                                             | warning                 |
    | defines a template where both the name and source are complex expressions | should not be displayed |
    | defines a template where name and source are both simple expressions      | should not be displayed |
    | defines a template where name is a complex expression                     | should not be displayed |
    | infers a template with an expression                                      | should not be displayed |
    | refers to a hidden template                                               | should not be displayed |
    | refers to a local template                                                | should not be displayed |
    | refers to a missing template                                              | should be displayed     |
    | refers to a template in a subdirectory                                    | should not be displayed |
    | refers to a template                                                      | should not be displayed |
    | refers to a template with an expression                                   | should not be displayed |
    | refers to a template without an erb extension                             | should not be displayed |
    | uses a missing inferred template                                          | should be displayed     |
    | uses an inferred template                                                 | should not be displayed |
    | uses a template from another cookbook                                     | should not be displayed |
```

that huge block is a big loop, and if you unroll it the first example is:

```
    Given a cookbook recipe that defines a template where both the name and source are complex expressions
     When I check the cookbook
     Then the missing template source file(s) warning 033 should not be displayed
```

second example is:

```
    Given a cookbook recipe that defines a template where name and source are both simple expressions
     When I check the cookbook
     Then the missing template source file(s) warning 033 should not be displayed
```

Then you should be able to find a 'Given' and a 'Then' which matches those lines.

A lot of the 'Then' lines could most likely be cleaned up and just use this 'Then' definition:

https://github.com/acrmp/foodcritic/blob/master/features/step_definitions/cookbook_steps.rb#L2102-L2123

If you can use that instead of writing yet-another-then-one-off its better.

For the 'Given' lines there's a bunch of existing cases of setting up templates and files here:

https://github.com/acrmp/foodcritic/blob/master/features/step_definitions/cookbook_steps.rb#L529-L565

They all follow the pattern of writing a recipe with a single template resource and then generally writing a template file (unless the template file is missing).  There's one that is parameterized and handles the 'missing' case.  The others are special cases like dotfiles where it doesn't make sense to re-test for missing again.

So, yeah, I dunno exactly what concept you're missing but you're clearly fighting the tool a lot harder than you need to...

docwhat:
Do you want any help on this PR?

tas50:
@docwhat This just needs proper tests written and it can be merged.

tas50:
Going with #485 instead since it has the tests

PR #431, opened 2016-03-04
[WIP]: No large binaries/files in cookbooks!

jjasghar:
- This is the intial commit to check to make sure you are not shipping gems with your cookbooks

- No gems were harmed in the writing of this PR.

### DIFF
--- a/lib/foodcritic/rules.rb
+++ b/lib/foodcritic/rules.rb
@@ -285,7 +285,7 @@ rule 'FC019', 'Access node attributes in a consistent manner' do
     if asts.size > 1
       least_used = asts[types.min do |a, b|
         a[:count] <=> b[:count]
-      end[:access_type]]
+        file[:access_type]]
       least_used.map do |file|
         file[:ast].map do |ast|
           match(ast).merge(filename: file[:path])
@@ -293,7 +293,12 @@ rule 'FC019', 'Access node attributes in a consistent manner' do
       end
     end
   end
-end
+end  recipe do |ast|
+    find_resources(ast, type: 'remote_file').find_all do |download|
+      path = (resource_attribute(download, 'path') ||
+        resource_name(download)).to_s
+      path.start_with?('/tmp/')
+    end
 
 rule 'FC021', 'Resource condition in provider may not behave as expected' do
   tags %w(correctness lwrp)
@@ -879,5 +884,12 @@ rule 'FC065', 'Ensure source_url is set in metadata' do
   end
   metadata do |ast, filename|
     [file_match(filename)] unless field(ast, 'source_url').any?
+  end
+end
+
+rule  'FC066', 'Ensure no gems are being shipped with cookbooks' do
+  tags %w(correctness annoyances chef12)
+  cookbook do |path|
+    [file_match(path.join(filename, '.gem'))]
   end
 end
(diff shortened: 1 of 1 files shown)


coderanger:
This seems oddly specific? Is this a common problem I'm not aware of?

doubt72:
Needs more cowbell.

jjasghar:
@coderanger Actually yes. In my travels with chef i've seen people try to ship gems with cookbooks. To be honest, this could any real executable; but we should start with gems.

coderanger:
Why would shipping a gem with a `cookbook_file` be wrong? Or at least any more wrong than any other type of package.

jjasghar:
I've seen people try to ship gems in `files/default` and wonder why syncs would take a long time. Ideally you'd have your gems in a shared location and the code would pull it down.

coderanger:
Sure, but that's at best a "best practice" not a lint check. And there are lots of good reasons to do it, just like there are reasons to ship debs and RPMs in a cookbook sometimes.

coderanger:
If you wanted to make a general check for this "best practice" it should probably be based on file size anyway, not extension.

tas50:
This should really be generalized to "don't ship large files in your cookbook" and we can decide on some large file size we think is evil (1 meg or so).  I've seen tons of people ship debs/wars/rpms in their cookbooks.  It is something we should discourage, but in a more general way.

coderanger:
Yeah, I would support either a per-file or total size threshold (or maybe both?) a lot more than this. This check feels like it would just leave a user confused.

cheeseplus:
Also shipping gems is totally a cool thing to do like for network devices where that is the best delivery mechanism.

jjasghar:
Having a check for a binary over a meg seems more reasonable. Not sure how to make this happen, but maybe that's more the spirit of what I'm trying to do here.

xenolinguist:
:+1: to changing this to a file size check.

jjasghar:
So if we move this to a file size check; do we have an idea "where the chef server starts erroring" at too big of a binary?

tas50:
It's not that we're breaking the Chef server we're just mis-using Chef, which is a perfectly acceptable thing to alert on in Foodcritic.  The Chef server is not an artifact store for releasing your application. That should be done in S3 or Artifactory/Nexus.  I think setting a 1meg limit is perfectly reasonable and if people really want to keep pushing binaries in their cookbooks they can always turn it off.

cheeseplus:
+1 to @tas50 point - I just didn't like us singling out gems as those are, in my exp, the most valid (ab)use

coderanger:
1M per file and 10M overall cookbook size seem like a reasonable place to start for a linter. If we had RubyCop-ish levels it might be nice to have both warning and error thresholds but alas :)

cheeseplus:
:+1:

jjasghar:
Awesome, so obviously this PR (code) is completely wrong now. Any chance I can ask some pairing time with someone to make this happen? I'm betting we could get this done in under an hour.

cheeseplus:
@jjasghar no foodcritic experience but I'd be up for pairing

tas50:
@jjasghar I'm going to close this out since the no large binaries check would be an entirely different thing.

PR #434, opened 2016-03-09
Rake.last_comment was depreciated in Rake11.

gkuchta:
See https://github.com/ruby/rake/commit/e76242ce7ef94568399a50b69bda4b723dab7c75

### DIFF
--- a/lib/foodcritic/rake_task.rb
+++ b/lib/foodcritic/rake_task.rb
@@ -16,7 +16,7 @@ module FoodCritic
 
 
       def define
-        desc 'Lint Chef cookbooks' unless ::Rake.application.last_comment
+        desc 'Lint Chef cookbooks' unless ::Rake.application.last_description
         task(name) do
           puts "Starting Foodcritic linting..."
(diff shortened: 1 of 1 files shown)


rcyrus:
:+1:

lib/foodcritic/version.rb
```diff
@@ -1,4 +1,4 @@
 module FoodCritic
   # The current version of foodcritic
-  VERSION = '6.0.1'
+  VERSION = '6.0.2'
```
tas50: Can you remove this.  We'll handle the version bump when we're ready for a release

gkuchta: @tas50 sure thing. i've removed the version bump.

tas50:
@gkuchta Can you rebase this down to a single commit too so the version doesn't bump in git history?

Thanks,
Tim

gkuchta:
@tas50 sure thing. i rebase'd off 7d512af31aebbf978f1be680cc519b84cef6af30 -- looking cleaner now?

PR #452, opened 2016-04-26
Add 12.9.38 metadata and make it the default chef version

tas50:
This brings in all the new resource attributes that shipped in 12.4.1 - 12.9.38


lamont-granquist:
👍

Issue #420, opened 2016-01-19
exclude_paths defaults differ between CLI and Rake:LintTask


err0:
The rake task interface has spec, test and features excluded by default.  the CLI interface does not. 
https://github.com/acrmp/foodcritic/blob/master/lib/foodcritic/rake_task.rb#L35

This inconsistency  should at a minimum be documented, if not made the same.

tas50:
This was fixed in FC 9.0

PR #462, opened 2016-06-30
More sane default CLI values

tas50:
1) Turn the progress on by default.  Foodcritic doing nothing by default is just plain odd. If someone wants silent output they should specify so
2) Skip files in test, spec, and features by default.  We're skipping these in the Rake task currently and test cookbooks almost always cause false alarms in FC. This will speed up the runs and improve user workflow.
3) Fail on any tag by default. It's odd that out of the box we show failures, but don't throw proper exit codes for them. This probably leads to a lot of incorrectly configured test jobs. I know I've done it wrong in the past (a few times). If someone wants to limit what they fail on they should specify it since that's flow is out of the norm.

### DIFF
--- a/lib/foodcritic/command_line.rb
+++ b/lib/foodcritic/command_line.rb
@@ -8,13 +8,14 @@ module FoodCritic
       @args = args
       @original_args = args.dup
       @options = {
-        fail_tags: [],
+        fail_tags: ['any'],
         tags: [],
         include_rules: [],
         cookbook_paths: [],
         role_paths: [],
         environment_paths: [],
-        exclude_paths: [],
+        exclude_paths: ['test/**/*', 'spec/**/*', 'features/**/*'],
+        progress: true
       }
       @parser = OptionParser.new do |opts|
         opts.banner = "foodcritic [cookbook_paths]"
@@ -27,9 +28,9 @@ module FoodCritic
                 "List all enabled rules and their descriptions.") do |t|
           @options[:list] = t
         end
-        opts.on("-f", "--epic-fail TAGS",
-                "Fail the build based on tags. Use 'any' to fail "\
-                "on all warnings.") do |t|
+        opts.on('-f', '--epic-fail TAGS',
+                "Fail the build based on tags. Default of 'any' to fail "\
+                'on all warnings.') do |t|
           @options[:fail_tags] << t
         end
         opts.on("-c", "--chef-version VERSION",
@@ -59,9 +60,9 @@ module FoodCritic
                 "foodcritic/rules/**/*.rb") do |g|
           @options[:search_gems] = true
         end
-        opts.on("-P", "--progress",
-                "Show progress of files being checked") do
-          @options[:progress] = true
+        opts.on('-P', '--[no-]progress',
+                'Show progress of files being checked. default:  true') do |q|
+          @options[:progress] = q
         end
         opts.on("-R", "--role-path PATH",
                 "Role path(s) to check.") do |r|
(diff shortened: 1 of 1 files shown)


tas50:
Fixes #420

lib/foodcritic/command_line.rb
```diff
@@ -26,9 +27,9 @@ def initialize(args)
         opts.on('-l', '--list',
                 'List all enabled rules and their descriptions.') do |t|
           @options[:list] = t
-        end                
+        end
         opts.on('-f', '--epic-fail TAGS',
-                "Fail the build based on tags. Use 'any' to fail "\
+                "Fail the build based on tags. Default of 'any' to fail "\
```
coderanger: I feel like this whole message needs to be tweaked.

lib/foodcritic/command_line.rb
```diff
@@ -76,7 +77,8 @@ def initialize(args)
           @options[:version] = true
         end
         opts.on('-X', '--exclude PATH',
-                'Exclude path(s) from being linted.') do |e|
+                'Exclude path(s) from being linted. Default test/**/*,'\
+                'spec/**/*,features/**/*') do |e|
```
coderanger: Given how this works, it looks like it might not be possible to remove the default? We should probably add something for that.

coderanger:
Will have to be a major rev but 👍 overall.

jaymzh:
I'm not sure skipping specs is a good call. It's the opposite of your third point (forcing more correctness). I'm down with the rest though.

lib/foodcritic/command_line.rb
tas50: Unfortunately in it's current state you'd have to pass some bogus dir via the CLI to override this.  Something like -X none would do it. That's less than ideal and I'm open up to a better solution.

tas50:
@jaymzh The specs dir is already skipped if you run the Rake task where this same set of directory exclusions is already set. I'm not really concerned about the specs, but when FC picks up test cookbooks is makes the app less than useful.  I'd hope that a test cookbook is well written, but some rules just don't make sense all the time.  Things like issues_url or source_url have no place in a test cookbook and should trigger FC.  That's my real goal here.

lib/foodcritic/command_line.rb
coderanger: Even that would just add `none` to the existing list I think. At a minimum we should probably make it so passing any option skips the defaults, that would at least make the workaround you mention an option.

charlesjohnson:
Should we also skip .delivery by default? I'm inclined to leave build cookbooks out of this whole thing, and let them have their own checking, but I can go either way.

tas50:
Delivery should also be excluded

tyler-ball:
As a follow up to this, we should change https://github.com/chef/chef-dk/blob/master/lib/chef-dk/skeletons/code_generator/files/default/delivery-project.toml#L20 to not add `--exclude spec` by default. If we continue doing that we will basically ignore the new defaults here.

This only affects the `delivery local` command.

tyler-ball:
I also think we should be running foodcritic on `.delivery/build-cookbook`. It is also a cookbook, so why shouldn't we be testing it?

tas50:
As time goes on we're adding more rules to foodcritic around sharing the cookbooks.  Test cookbooks and build cookbooks lack a lot of those fields.  Thinks like long_description, source_url and issues_url just don't make sense in test / build cookbooks. They were created as rules for the meat of the cookbook. If we're running tests against those cookbooks we're going to have to start weeding out tests for those cookbooks. At the very least we'd need a way to exclude rule tags for certain "sub" cookbooks within the cookbook. Overall it seems like a lot of work for very little gain.

PR #464, opened 2016-06-30
Remove chef version support for Chef 0.7, 0.8, 0.9, and 0.10

tas50:
We should assume that anyone using modern Foodcritic is trying to test Chef cookbooks written at least for Chef 11.  This greatly reduces the size of the Foodcritic gem and most likely impact 0 users.


lamont-granquist:
👍

PR #465, opened 2016-06-30
Remove support for Ruby 2.0

tas50:
Ruby 2.0 is EOL. We should only support currently Ruby releases. I'd remove support for 2.1 as well, but that's what ChefDK ships with so we can't.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: false
 language: ruby
 cache: bundler
 rvm:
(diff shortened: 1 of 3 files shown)


lamont-granquist:
👍

lamont-granquist:
rack obviously needs pinning here as well

tas50:
That comes along for free when chef 12.12 ships.

PR #478, opened 2016-08-17
Add missing assigment attributes

ofir-petrushka:
Add a few missing attributes to fix rule FC047 false positives.

Related issue
https://github.com/acrmp/foodcritic/issues/477


lamont-granquist:
well that explains why people kept complaining after i thought i fixed this...

PR #485, opened 2016-09-18
Support Chef RFC 17 compliant templates

### LINKED ISSUE #476: FC033: does not recognize templates in non-default dir
We like to separate our templates into specific folders if we write a LWRPs.

We get:
`FC033: Missing template: ./resources/logentries_follow.rb:16` 
for 

```
template app_name do
  path "/etc/le/conf.d/#{app_name}.conf"
  source 'logentries/base.conf.erb'
```

Obviously the template is there or our recipes wouldn't be working. Maybe this is not the standard? Issue or not? 

Thanks!

hagihala:
This adds support for the template source files placed in the root of the templates directory.
This behavior is defined in [Chef RFC 17](https://github.com/chef/chef-rfc/blob/master/rfc017-file-specificity.md) and was introduced in Chef 12.
It is related to #425, but will just fix #476.


hagihala:
Test with Ruby 2.1 fails,  maybe because chef and chef-zero have dropped Ruby 2.1 support.

yeoldegrove:
It would be great to merge this as it fixes the anoying FC033 related bugs.

PR #489, opened 2016-10-03
Clarify exclude path instructions

### LINKED ISSUE #488: --help output for --exclude is confusing
I was seeing foodcritic 8.0.0 descend into test for some of my cookbooks even when I explicitly called `--exclude cookbookname/test`.

It was unclear from the --help output that I needed to `--exclude test` instead of `--exclude cookbook_name/test`

unixorn:
When I was beginning to use foodcritic, it was detecting the test subdirectory of some of our cookbooks as another cookbook and throwing false FC011, FC031 and FC045 errors.

When I was trying to exclude the test directory tree, I read the instructions as needing to specify an absolute path, not one relative to the cookbook directory.

This PR spells that out in the `--help` output.

Closes #488

### DIFF
--- a/lib/foodcritic/command_line.rb
+++ b/lib/foodcritic/command_line.rb
@@ -76,7 +76,7 @@ module FoodCritic
           @options[:version] = true
         end
         opts.on("-X", "--exclude PATH",
-                "Exclude path(s) from being linted.") do |e|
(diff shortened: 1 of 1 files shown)


tas50:
Certainly worth adding there. Thanks

unixorn:
You're welcome. That was a smoking fast merge :-)

PR #507, opened 2017-01-25
Test on Ruby 2.4 and fix failing tests

tas50:

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,7 +2,7 @@ require "foodcritic/version"
 require "bundler"
 require "rake/testtask"
 
-task :default => [:man, :install, :rubocop, :test, :features]
+task :default => [:man, :install, :test, :features, :rubocop]
 
 Bundler.setup
 Bundler::GemHelper.install_tasks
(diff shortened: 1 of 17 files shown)


Rakefile
```diff
@@ -2,7 +2,7 @@ require "foodcritic/version"
 require "bundler"
 require "rake/testtask"
 
-task :default => [:man, :install, :rubocop, :test, :features]
+task :default => [:man, :install, :test, :features, :rubocop]
```
josephholsten: usually I sort these by how long it takes to run, so I'm a bit confused by this change

tas50: This is something lamont brought up today and I was running into it. Style is the thing you care about the least and you often fail tests there first, which stops testing. This way you see your failing specs first and you can clean up your style later. It avoids multiple commits to style.

tas50:
I've removed the tests that check for the exist status as these all fail now. Our default behavior is now to fail on any failing tag. If we generate a cookbook that lacks basic things like metadata it will ALWAYS fail a tag and thus always exit non-0. There's no way to test this in the current design of the tests / app.

Rakefile
lamont-granquist: yeah it annoys the crap out of me to have rubocop fail first because fixing that is usually mindless.

PR #511, opened 2017-02-09
Require Chef 12.6 or later

tas50:
Removes the metadata files that allowed checking against pre 12.6 cookbook syntax. This has very little impact as the current metadata is 12.18.X

This nukes about 100k from the gem so that's cool too


tas50:
Closing to instead remove the metadata altogether

PR #512, opened 2017-02-09
Remove FC003 from Foodcritic

tas50:
It no longer makes sense now that solo uses local mode and search works with that.


features/support/command_helpers.rb
```diff
@@ -16,7 +16,7 @@ def assertions
     WARNINGS = {
       "FC001" => "Use strings in preference to symbols to access node attributes",
       "FC002" => "Avoid string interpolation where not required",
-      "FC003" => "Check whether you are running with chef server before using server-specific features",
+      # FC003 was yanked and is considered reserved, do not reuse it
```
iennae: yanked might not make sense to international audience. maybe "FC003 has been deprecated and is considered reserved.."

lib/foodcritic/rules.rb
```diff
@@ -26,16 +26,7 @@
...
-     "Check whether you are running with chef server before using"\
-     " server-specific features" do
-  tags %w{portability solo}
-  recipe do |ast, filename|
-    unless checks_for_chef_solo?(ast) || chef_solo_search_supported?(filename)
-      searches(ast)
-    end
-  end
-end
+# FC003 was yanked and the number should not be reused
```
iennae: # FC003 was a check for Chef server specific features that 
# did not interoperate with the implementation of chef-solo.
# It has been deprecated, and should not be reused.

iennae: Making it a comment makes it think I meant this as a giant header. doh.

features/support/command_helpers.rb
tas50: This is just copy pasta from the last FC rule we deprecated. The actual docs for FC word is differently. This is just a code comment to prevent devs from reusing the number

lib/foodcritic/rules.rb
tas50: Same as above these aren't actually seen by end users so the wording isn't the important part. That's for the PR against foodcritic-site

PR #514, opened 2017-03-05
Retire rules that are better suited for Chefspec FC009/FC037/FC038

tas50:
FC009/FC037/FC038 detect typos/mistakes in code by dumping every possible method for each version of chef and then comparing cookbook code against that list. This is flawed for many reasons:

Requires constantly updating chef metadata
Doesn’t work if you’re using an old version of foodcritic but new functionality from a new chef release
Fails on the windows resources that are in the cookbook

This test made sense when it was developed, but at this point a converge of the recipe using Chefspec is far more likely to find the issue we’re hoping to discover here


tas50:
So this actually requires a lot of fundamental reworking that I'm not prepared to do now (or maybe ever). This might just stick around until we one day port these rules all over to rubocop

PR #519, opened 2017-03-16
Fix FC045 to not fail if not metadata.rb file exists

tas50:
So it turns out all our metadata checks don't work on the Supermarket if you uploaded using stove which strips off the metadata.json. Most just silently skip, but FC045 hard fails, which is not the desired way to do this. Instead lets just try to return the name value starting with metadata.rb and failing back to metadata.json. We'll raise if that doesn't return anything. Assuming people like this pattern we can update the other metadata checks to use the same pattern.


tas50:
This won't work due to the differences in the JSON data vs. metadata.rb data. It's just too much magic to make work

PR #520, opened 2017-03-16
Add a more robust base path helper to the API

tas50:
The current methods of discovering the base path are flawed as they fail if the cookbook is within a directory named the same as a cookbook directory (templates, recipes, files, etc). This gives us an actual helper for finding the base path of a cookbook. We can migrate existing api calls and rules over to this in the future.

### DIFF
--- a/lib/foodcritic/api.rb
+++ b/lib/foodcritic/api.rb
@@ -89,7 +89,7 @@ module FoodCritic
 
       # get list of items in the dir and intersect with metadata array.
       # until we get an interfact (we have a metadata) walk up the dir structure
-      until (Dir.entries(file) & %w(metadata.rb metadata.json)).any?
+      until (Dir.entries(file) & %w{metadata.rb metadata.json}).any?
         file = File.absolute_path(File.dirname(file.to_s))
       end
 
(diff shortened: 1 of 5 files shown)


lib/foodcritic/api.rb
```diff
@@ -77,6 +77,20 @@ def chef_solo_search_supported?(recipe_path)
...
 
+    # The absolute path of a cookbook from the specified file.
+    def cookbook_base_path(file)
+      file = File.expand_path(file) # make sure we get an absolute path
+      file = File.dirname(file) unless File.directory?(file) # get the dir only
+
+      # get list of items in the dir and intersect with metadata array.
+      # until we get an interfact (we have a metadata) walk up the dir structure
+      until (Dir.entries(file) & %w(metadata.rb metadata.json)).any?
+        file = File.absolute_path(File.dirname(file.to_s))
```
josephholsten: I think we can skip the `File.absolute_path` since we've run `File.expand_path` on its descendant

lib/foodcritic/api.rb
```diff
@@ -78,6 +78,11 @@ def chef_solo_search_supported?(recipe_path)
     end
 
     # The absolute path of a cookbook from the specified file.
+    #
+    # @author Tim Smith - tsmith@chef.io
+    # @since 11.0
+    # @param file [String, Pathname] relative or absolute path to a file in the cookbook
```
josephholsten: I think it probably works with a `Pathname` but we should test to be sure ;-)

tas50:
@josephholsten Fixed specs that haven't been running for a super long time and added a test of a few different scenarios

josephholsten:
¡super approval!

PR #531, opened 2017-03-31
Add FC070 to detect invalid platform supports in metadata

tas50:


lib/foodcritic/rules/fc070.rb
```diff
@@ -0,0 +1,59 @@
...
+          matches = true
+          break # we found a bad platform so stop looking
+        end
+      end
+    end
+
+    # see if the platform is a common mistake platform
+    unless matches
+      metadata_platforms.each do |plat|
+        if invalid_platforms.include?(plat)
```
robbkidd: Should we `.downcase` `plat` here to make the match case-insensitive?

tas50: Oh that reminds me I forgot to add a check for upcase platforms

PR #537, opened 2017-04-06
Testing harness overhaul

coderanger:
This started as just a new functional testing setup but has become a full revamp of all testing.


coderanger:
So if you want to run just one type of tests you can now do `rake 'spec[unit]'` or `rake 'spec[functional]'`. The regression tests are passing though regenerating them would be good too to get better coverage. Minitest is banished, and a framework for replacing Cuke for functional tests is in place. I basically copy-pasta'd some of my rake and spec_helper bits, which gets some nice features like improved output and focus filtering (use `fdescribe` or `fit` to run only a specific set of tests when developing).

tas50:
This one is more than I can digest, but the idea seems to move us in the right direction even if we need to cleanup some things later.

PR #541, opened 2017-04-08
Add FC071 to check that LICENSE file is included

tas50:
Stolen from https://github.com/sous-chefs/sc-foodcritic-rules/blob/master/rules.rb


tas50:
@coderanger Does Github recognize COPYING. That's pretty common in older projects, but Github keys off LICENSE with some cool metadata now.

tas50:
Yep they're using a gem that only looks for the license file, which is the goal here. If people really want copying that can be added, but I suspect that sourceforge crowd and the chef crowd don't overlap much.

coderanger:
:+1:

PR #542, opened 2017-04-08
Add FC072 for alerting on the attribute keyword in metadata

tas50:


tas50:
@coderanger

PR #543, opened 2017-04-08
Improved regression tests

coderanger:
Also minor tweak to the README.


tas50:
So much easier to read. Thanks

PR #544, opened 2017-04-08
Convert FC 11,12,16,28 and 42 to rspec

tas50:
Get a few more of these converted before fixing some bugs that will require expanding the testing


tas50:
@coderanger

PR #545, opened 2017-04-08
FC008 should detect ChefDK boilerplate metadata

tas50:
We only detect the knife cookbook create boilerplate and this command has been removed from Chef 13. Detect both sets of boilerplate.


tas50:
@coderanger

PR #549, opened 2017-04-09
Root alias stuffs

coderanger:
Includes a fix (with test) for scanning root alias files and a new rule to detect the case of alias collision.


lib/foodcritic/rules/fc073.rb
```diff
@@ -0,0 +1,17 @@
+rule "FC073", "Root alias file shadowing non-alias file" do
+  tags %w{correctness}
```
tas50: This should have the chef13 tag as well

tas50:
@coderanger Make sure to write a page for it on the foodcritic-site repo now

PR #551, opened 2017-04-09
Convert more cucumber to rspec

tas50:


tas50:
@coderanger

spec/functional/fc061_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    metadata_file('version "1.2.3"')
+    it { is_expected.not_to violate_rule("FC061") }
+  end
+
+  context "with a cookbook with metadata that includes the version keyword and a valid x.y version string" do
+    metadata_file("version '1.2'")
+    it { is_expected.not_to violate_rule("FC061") }
+  end
+
+  # this should be the case, but FC061 has a bug
```
coderanger: Rather than commenting it, you could add this as a pending test so we remember to fix it later :)

tas50: Yep that would make a lot more sense

PR #555, opened 2017-04-09
Add FC075 to detect node.save

tas50:
node.save is something you should only do if you REALLY know what you’re doing and are entirely sure actually you want a node.save to occur. Most people don’t need it and we really don’t want community cookbooks randomly saving the node (some do)

### DIFF
--- a/lib/foodcritic/rules/fc075.rb
+++ b/lib/foodcritic/rules/fc075.rb
@@ -1,0 +1,9 @@
+rule "FC075", "Cookbook uses node.save to save data to the chef-server mid-run" do
+  tags %w{correctness}
+  def node_saves(ast)
+    ast.xpath('//call[(vcall|var_ref)/ident/@value="node"]
+    [ident/@value="save"]')
+  end
+  recipe { |ast| node_saves(ast) }
+  library { |ast| node_saves(ast) }
+end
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -68,6 +68,10 @@ def resource_file(*args, &block)
       file("resources/my_resource.rb", *args, &block)
     end
 
+    def libary_file(*args, &block)
+      file("resources/helper.rb", *args, &block)
```
coderanger: That's not a library file?

spec/functional/fc075_spec.rb
```diff
@@ -0,0 +1,36 @@
+require "spec_helper"
+
+describe "FC075" do
```
coderanger: Might be good to have a negative test for `node['save']` or whatevers.

lib/foodcritic/rules/fc075.rb
```diff
@@ -0,0 +1,9 @@
+rule "FC075", "Cookbook uses node.save to save data to the chef-server mid-run" do
```
coderanger: Maybe "to save partial data" to make it clearer why this is bad?

coderanger:
+1 for the idea, just need to fix the spec helper :)

spec/spec_helper.rb
tas50: and I should be done for the night

PR #588, opened 2017-04-10
Run the regression testing against more complex cookbooks

tas50:
Remove a few incredibly simple cookbooks and replace them with older releases of far more complex cookbooks


spec/regression/regression_spec.rb
```diff
@@ -1,7 +1,7 @@
 require "spec_helper"
 
 describe "regression test" do
-  command("#{File.expand_path("../../../bin/foodcritic", __FILE__)} .", allow_error: true)
+  command("#{File.expand_path("../../../bin/foodcritic --tags any", __FILE__)} .", allow_error: true)
```
coderanger: The args should be outside the expand path :)

PR #590, opened 2017-04-10
Fix a regression when running Foodcritic as a Rake task

RoboticCheese:
The Output classes' initializers now [require](https://github.com/acrmp/foodcritic/commit/3c057cb857939b6761effb386fe0b30e9270add2)
an output stream argument.

Signed-off-by: Jonathan Hartman <j@p4nt5.com>


tas50:
Shipped at 10.3.1

PR #593, opened 2017-04-11
new rule for Supermarket to check for open source license

robbkidd:
Placed in only the supermarket category. Cookbooks shared on the public Supermarket are expected to be shared under an open source license. This rule uses the [list of licenses from SPDX](https://spdx.org/licenses/) that are identified as OSI-approved. Ideally, this rule will be used to meet the [license quality metric](https://github.com/chef-cookbooks/cookbook-quality-metrics/blob/35bddc5af54eb8478b2307f7b6d231f03cf61ae9/quality-metrics/qm-003-license.md) with a foodcritic run.


robbkidd:
Reading up more on SPDX license matching, I note that I am being stricter here (as in [FC069](https://github.com/acrmp/foodcritic/blob/15e85b5dd8ec577f0427fdeee229e99c9709c995/lib/foodcritic/rules/fc069.rb#L360)) on the string match than SPDX calls for (e.g. the leniency granted in [the SPDX matching guidelines](https://spdx.org/spdx-license-list/matching-guidelines) around Whitespace, Capitalization, and Punctuation).

szymonpk:
Overall :+1:. As discussed on Slack with @robbkidd. We just should keep in mind, that any future rules that aren't specific to open source/public supermarket should have at least one additional (other than supermarket) category. Some people may want to keep all metrics quality checks with a possibility of having `all rights reserved` license, which isn't allowed by this rule.

lib/foodcritic/rules/fc078.rb
```diff
@@ -0,0 +1,135 @@
+rule "FC078", "Ensure cookbook shared under an OSI-approved open source license" do
+  tags %w{supermarket}
```
tas50: I think we're going to want another tag here so people can disable open source type checks in their internal CI systems. Perhaps 'community' or 'opensource'

robbkidd: That also makes sense, but has me wondering what the meaning of the `supermarket` tag is. I suppose a `community` tag could be applied to rules that map to [cookbook quality metrics](https://github.com/chef-cookbooks/cookbook-quality-metrics/blob/master/SMQM.md).

robbkidd: Now that I've experimented with `foodcritic` and `-t ~tag`, I see that the `~` wins regardless of what other tags are enabled. So, I suppose `opensource` might be better so that it can be the only tag on a rule that is only related to open source.

For example, `foodcritic -t metadata -t ~supermarket` turned off _any_ `metadata` rule that also had `supermarket` on it. 😞

tas50: I think the scenario is an org with a private supermarket. They want to ensure people use issues_url, but they don't care about open source licenses

PR #595, opened 2017-04-12
Update the metadata rules to use "deprecated" vs "unimplemented"

tas50:
Unimplemented doesn't convey the reason you should remove these from
your code. If you're on 11, 12, or 13 you want these gone


tas50:
Ping @coderanger

PR #596, opened 2017-04-13
Ignore case when checking for "All Rights Reserved" in FC071

jyaworski:
We should not care about the case of the proprietary license, due to there not being a SPDX standard.


tas50:
@jyaworski There's some test failures being caused by this change. Also can you create a functional spec that tests the lowercase value not throwing an error for the user? Let me know if you need some help with that.

tas50:
I opened https://github.com/acrmp/foodcritic/pull/607 which adds a .to_s which prevents failures when there's no license. Those were breaking existing specs. I also added a spec.

PR #597, opened 2017-04-14
Use the given cookbook path in FC071

dduponchel:
This fix the rule FC071 when used with a cookbook path:

    foodcritic -t FC071 my-cookbook

The file `metadata.rb` doesn't exist in the current folder but it does
in `my-cookbook`.

### DIFF
--- a/lib/foodcritic/rules/fc071.rb
+++ b/lib/foodcritic/rules/fc071.rb
@@ -1,7 +1,7 @@
 rule "FC071", "Missing LICENSE file" do
   tags %w{style license}
   cookbook do |path|
-    unless ::File.exist?("metadata.rb") && field_value(read_ast("metadata.rb"), "license") == "All Rights Reserved"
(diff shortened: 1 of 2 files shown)


coderanger:
Would be super cool if you made a regression test :)

dduponchel:
That would mean a new entry `spec/regression/cookbooks.txt`, right ? This test is about proprietary cookbooks that are likely to be closed source (like mine). I searched [on the chef-cookbooks organization](https://github.com/search?p=1&q=org%3Achef-cookbooks+%27license+All+Rights+Reserved%27&type=Code&utf8=%E2%9C%93) but the only matches are in examples or tests folders.

Do you have any pointers to create a meaningful test ?

coderanger:
Ahh no, I just meant a functional test that catch the bug.

dduponchel:
I added two tests. Are you ok with them ?

tas50:
@dduponchel Thanks for the tests. I'll get a release cut with this change.

dduponchel:
Thanks !

PR #599, opened 2017-04-14
Update FC028 to detect bad platform_family usage as well

tas50:


tas50:
Ping @coderanger

PR #605, opened 2017-04-16
Strip leading whitespace on platforms to fix FC070

tas50:
This fails due to the way the AST is parsed, but is totally valid Ruby. Our helper for determining the supported platforms in the metadata needed to strip leading whitespace.

    %w(
      amazon
      centos
      debian
    ).each do |os|
      supports os
    end

### DIFF
--- a/lib/foodcritic/api.rb
+++ b/lib/foodcritic/api.rb
@@ -393,7 +393,7 @@ module FoodCritic
       platforms.map do |platform|
         versions = platform.xpath('ancestor::args_add[position() > 1]/
 	  string_literal/descendant::tstring_content/@value').map { |v| v.to_s }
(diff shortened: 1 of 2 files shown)


tas50:
Fixed https://github.com/acrmp/foodcritic/issues/603

tas50:
So the trailing whitespace isn't an issue, but the parsing logic is pretty flawed.

```
[1] pry(#<FoodCritic::RuleDsl>)> supported_platforms(ast).map { |x| x[:platform] }
=> ["#", "amazon", "trailing", "whitespace"]
```

tas50:
I'm going to see if there's a way to fix that issue in another PR. This gets 1 issue out of the way

coderanger:
:+1:

PR #612, opened 2017-04-21
Add FC080 to detect user resources using supports property

tas50:


spec/functional/fc080_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+    EOF
+    it { is_expected.to violate_rule("FC080") }
+  end
+
+  context "with a cookbook with a recipe that includes a user resource not using supports" do
+    recipe_file <<-EOF
+      user "betty" do
+        action :create
+        manage_home: true,
```
coderanger: Should should be `manage_home true`.

tas50: true

PR #614, opened 2017-04-22
Add FC082: Node.set used to set node attributes

tas50:


tas50:
Do people do the read with that?

coderanger:
100% chance someone somewhere is doing `foo = node.set['foo']`.

tas50:
ahh that usage makes sense

PR #615, opened 2017-04-22
Add FC083: Execute resource using 'path' property

tas50:


lib/foodcritic/rules/fc083.rb
```diff
@@ -0,0 +1,10 @@
+rule "FC083", "Execute resource using 'path' property" do
+  tags %w{deprecated chef12}
+  recipe do |ast|
+    matches = []
```
coderanger: I think you can do this more compressed:

```ruby
find_resources(ast, type: "execute").xpath('(.//command|.//fcall)[ident/@value="path"]')
```

That should be the whole check.

PR #616, opened 2017-04-22
Lots of nonsense

coderanger:
So this branch got kind of out of hand with the yak shaving but some highlights:

1. Rework some of the XPath expressions to be more flexible, notably `field` now groks calls with parens.
2. Remove a bunch of secondary xpath calls that duplicate features in the API mixin.
3. Tweak FC069 to be a bit more general with the "All Rights Reserved" check since people keep complaining.
4. Convert some more cukes because it was easier to debug the failures as specs.
5. Fix an RSpec warning.


coderanger:
@tas50 Ready for review :)

PR #617, opened 2017-04-24
Add FC084: Deprecated Chef::REST class used

tas50:

### DIFF
--- a/lib/foodcritic/rules/fc084.rb
+++ b/lib/foodcritic/rules/fc084.rb
@@ -1,0 +1,9 @@
+rule "FC084", "Deprecated Chef::REST class used" do
+  tags %w{chef13 deprecated}
+  def rest(ast)
+    ast.xpath('//const_path_ref/const[@value="REST"]/..//const[@value="Chef"]/../../..')
+  end
+  recipe { |ast| rest(ast) }
+  library { |ast| rest(ast) }
+
+end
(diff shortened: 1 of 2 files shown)


lib/foodcritic/rules/fc084.rb
```diff
@@ -0,0 +1,9 @@
+rule "FC084", "Deprecated Chef::REST class used" do
+  tags %w{chef13 deprecated}
+  def rest(ast)
+    ast.xpath('//const_path_ref/const[@value="REST"]/..//const[@value="Chef"]/../../..')
+  end
+  recipe { |ast| rest(ast) }
+  library { |ast| rest(ast) }
+
```
coderanger: Robucop will probably be sad about this blank line.

lib/foodcritic/rules/fc084.rb
```diff
@@ -0,0 +1,9 @@
+rule "FC084", "Deprecated Chef::REST class used" do
+  tags %w{chef13 deprecated}
+  def rest(ast)
+    ast.xpath('//const_path_ref/const[@value="REST"]/..//const[@value="Chef"]/../../..')
```
coderanger: Might be easier to say something like this `//const_path_ref[const[@value="Chef"]][const[@value="REST"]]`? Not sure that's exactly equivalent though.

lib/foodcritic/rules/fc084.rb
tas50: loves it clearly

lib/foodcritic/rules/fc084.rb
tas50: That doesn't detect it

PR #618, opened 2017-04-24
Add FC085: Resource using new_resource.updated_by_last_action to converge resource

tas50:
There's so much misuse of this pattern in our cookbooks. With use_inline_resources turned on by default now there's no reason to ever use it. We need to point people to simple pure chef resource based converges or to converge_by when they're writing their own ruby code in resources


tas50:
So this currently does not detect @new_resource usage. That's a known limitation

robbkidd:
Are there helpful words somewhere to point a person towards when this rule is tripped?

tas50:
@robbkidd That will go on the foodcritic-site repo. We'll have a "dont do this" / "do this" code example

PR #622, opened 2017-04-25
Be more clear these things are bad because they’re deprecated

tas50:


tas50:
We sorta have a mix of rule names already based on who the author is. Do we alert on the positive of what you should do, or the negative of what you did? We certainly could use some sort of standard. It's a bit of a mess.

Issue #568, opened 2017-04-10
Chef 13: Alert if someone uses Chef::Platform.set  in a resource


tas50:
Chef::Platform.set has been removed from Chef 13 and we should alert if someone uses it

https://github.com/chef/chef/pull/5914

tas50:
This is FC087

PR #624, opened 2017-04-27
FC087: Library maps provider with deprecated Chef::Platform.set

tas50:


tas50:
Resolves https://github.com/Foodcritic/foodcritic/issues/568

coderanger:
We should make it clear in the fc-site examples on how to port to using `provides` in the resource/provider with a filter condition.

tas50:
That should be in the deprecation docs site page. If it isn't we should add it there

PR #626, opened 2017-05-04
Add option to specify foodcritic rule file

dlukman:
I have added a very simple way to use a specific file with foodcritic rules instead of always looking for .foodcritic in the root of a cookbook.


tas50:
Thanks.

PR #634, opened 2017-05-22
Docker to Foodcritic

amalucelli:
Just a simple Docker image that helps the use of Foodcritic inside a CI tool without the need to download and build the latest version, for example.

I also added a basic info on how to use it.

If you guys think is relevant, we can also create a organization in [Docker Hub](https://hub.docker.com/) so people can pull the image without the need to build it, like `docker pull foodcritic/foodcritic`.


tas50:
Nice addition. Thanks

Issue #636, opened 2017-05-24
food critic should check for platform version format


bharathcp:
The `supports` platform version line of `metadata.rb` should be compliant to `x.y.z` or `x.y` format. Without this supermarket upload will fail with below error:
```
{"error_code"=>"INVALID_DATA", "error_messages"=>["Version constraint is not a valid Chef version constraint. Chef::Exceptions::InvalidCookbookVersion: '7' does not match 'x.y.z' or 'x.y'"]}
```
wrong format:
`supports 'centos', '>= 7'`
correct format:
`supports 'centos', '>= 7.3'`
There should be a rule that checks this.

tas50:
@bharathcp What command are you running to do the upload?

tas50:
Closing this out due to lack of response. The 2nd field in the version uploads really shouldn't be an issue and it's something that chef doesn't actually consume right now. If we're failing to upload that to Supermarket via stove or knife that's a bug there.

PR #637, opened 2017-05-24
added a rule for supports version numbers

bharathcp:


bharathcp:
fixes #636

tas50:
I'm going to close this out based on what @coderanger said. This would be really problematic if we implemented it.

PR #638, opened 2017-05-25
Use the improved spec helper method

tas50:


tas50:
Merging this since it's a simple test update

PR #641, opened 2017-06-07
Don’t run the open source tag by default

coderanger:
This disables FC078 by default under the reasoning that the majority of Chef developers are not releasing their code on Supermarket, and so we should make that the default here too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Foodcritic Changelog:
 
+## [11.2.0](https://github.com/acrmp/foodcritic/tree/v11.2.0) (...)
+
+[Full Changelog](https://github.com/acrmp/foodcritic/compare/v11.1.0...v11.2.0)
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Foodcritic Changelog:
 
+## [11.2.0](https://github.com/acrmp/foodcritic/tree/v11.2.0) (...)
+
+[Full Changelog](https://github.com/acrmp/foodcritic/compare/v11.1.0...v11.2.0)
+
+**Implemented enhancements:**
+
+- Disabled `FC078` by default to simply use of Foodcritic for non-community cookbook developers. To enable it again: `foodcritic -t any .`
```
tas50: We should not that we disabled the opensource tag which includes FC078 by default. There will probably be more things tagged with opensource in the future and this makes it sound like we made the default ~FC078

PR #651, opened 2017-07-10
Limit the files we ship in the gem

tas50:
Strip out the following files:

Gemfile
Rakefile
specs
features

This drops our gem from 269kb to 205k and the install size drops from 7.5M to 6.5M


thommay:
Yeah I think this will do ungood things to appbundling, too.

Issue #652, opened 2017-07-11
FC001 should not fire node.run_state[:foo]


tas50:
node.run_state[:foo] is valid. We should not alert on these.

tas50:
Fixed in Foodcritic 11.3

PR #653, opened 2017-07-12
FC001 should not alert on node.run_state[:foo]

tas50:


tas50:
Resolves https://github.com/Foodcritic/foodcritic/issues/652

PR #659, opened 2017-08-15
Warning no longer requires fail_tags as documented

bbaugher:


coderanger:
What does this actually fix?

bbaugher:
Its [documented](https://github.com/bbaugher/foodcritic/blob/5fe18099a7bea3fe8df325d19310e27f724c79d9/lib/foodcritic/domain.rb#L10-L11) that you don't have to provide the options when building a warning but this isn't true

coderanger:
I would fix the code comment instead, and also that's not documentation. That's a deeply internal API so no one outside of Foodcritic should be touching it.

PR #666, opened 2017-09-27
Add FC093 for generated README text

### LINKED ISSUE #658: Catch TODO in autogenerated README
When we use `chef generate cookbook`, we get an autogenerated `README.md` containing this line:

```
TODO: Enter the cookbook description here.
```

Could we make a rule to catch this? Like there are rules to catch autogenerated lines in `metadata.rb`.

gordonbondon:
Should fix #658

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -88,6 +88,10 @@ module FunctionalHelpers
       file("metadata.rb", *args, &block)
     end
 
+    def readme_file(*args, &block)
+      file("README.md", *args, &block)
+    end
+
     def included(klass)
       super
       klass.extend ClassMethods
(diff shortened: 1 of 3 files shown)


tas50:
Way cool. Thanks

tas50:
One more thing actually. Can you open a PR to https://github.com/Foodcritic/foodcritic-site so we can document the new rule.

spec/spec_helper.rb
```diff
@@ -88,6 +88,10 @@ def metadata_file(*args, &block)
       file("metadata.rb", *args, &block)
     end
 
+    def readme_file(*args, &block)
```
coderanger: This might be overkill for a dedicated helper :)

martinmosegaard:
Thanks @gordonbondon 👍

gordonbondon:
@tas50 done - Foodcritic/foodcritic-site#43

PR #675, opened 2017-10-12
FC090: Resource should not ignore failures.

johnbellone:
/cc @jonlives @coderanger @tas50 

This foodcritic rule fails if any resource sets [ignore_failure
true](https://docs.chef.io/resource_common.html#properties) property.

### DIFF
--- a/lib/foodcritic/rules/fc090.rb
+++ b/lib/foodcritic/rules/fc090.rb
@@ -1,6 +1,6 @@
 rule "FC090", "Resource should not ignore failures" do
   tags %w{correctness resource}
   recipe do |ast|
-    find_resources(ast).select { |resource| resource_attributes(resource)['ignore_failure'] }.compact
+    find_resources(ast).select { |resource| resource_attributes(resource)["ignore_failure"] }.compact
   end
 end
--- a/lib/foodcritic/rules/fc090.rb
+++ b/lib/foodcritic/rules/fc090.rb
@@ -1,0 +1,6 @@
+rule "FC090", "Resource should not ignore failures" do
+  tags %w{correctness resource}
+  recipe do |ast|
+    find_resources(ast).select { |resource| resource_attributes(resource)['ignore_failure'] }.compact
+  end
+end
--- a/lib/foodcritic/rules/fc090.rb
+++ b/lib/foodcritic/rules/fc090.rb
@@ -1,5 +1,5 @@
 rule "FC090", "Resource should not ignore failures" do
-  tags %w{correctness resource}
+  tags %w{resource}
   recipe do |ast|
     find_resources(ast).select { |resource| resource_attributes(resource)["ignore_failure"] }.compact
   end
--- a/spec/functional/fc090_spec.rb
+++ b/spec/functional/fc090_spec.rb
@@ -1,4 +1,4 @@
-require 'spec_helper'
+require "spec_helper"
 
 describe "FC090" do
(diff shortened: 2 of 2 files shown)


jonlives:
I'd like to hear @lamont-granquist's thoughts on this too but for the same reason as his comment in https://github.com/chef/chef-rfc/pull/279 I'm not sure this could be a correctness rule enabled by default because it'd effectively mean people couldn't use ignore_failure on purpose without having to disable the rule.

johnbellone:
@jonlives Ah, fair point!

lamont-granquist:
This is actually what I suggested, though.

It is somewhat naughty but not verboten to `ignore_failure`.  Putting a switch in Chef::Config opens up warfare between sites that enable it by default and community cookbooks that need to use it as a dirty hack.  Putting it in foodcritic means that those cookbooks that need to use it can disable the warning.

And we still support the `epic_fail` alias so should detect that here as well.

coderanger:
It's not quite as bad as the situation with a global since it would be per cookbook, but I'm worried this is a bit too opinionated for foodcritic since we don't have an info or warning level kind of thing. It's definitely right on the edge though given how rare legit uses are, so I defer to @tas50.

lamont-granquist:
My feeling is that this is not used much in community cookbooks and doesn't hit the level of other foodcritic rules that we've disabled because violating them was so common, but yeah, i'll defer to @tas50.

This also feels more like a thing that if I was talking to someone at ChefConf I'd tell them never to use this feature unless they really needed to hack around bugginess somewhere as a last resort -- while other foodcritic rules that we've disabled I would have had to be like "yeah you should prefer not to do that, but there's a pile of common occurrences when you'll need to know when to violate that rule..."

jonlives:
@tas50 you able to review this please?

tas50:
I'm going to post this same response to several issues since we have a theme going on here. 

Foodcritic is a pretty versatile cookbook validation tool. You can use it out of the box to get helpful tips on your cookbooks, you can write your own rules, and it's used with Supermarket to enforce quality metrics. Those are 3 distinct use cases and these use cases often conflict with each other. Things we think are code smell within community cookbooks are just fine for private cookbooks. Take for instance the issues we had recently with enforcement of open source licenses and SPDX compliant license strings in the metadata.rb file. That ONLY matters if you're uploading to the Supermarket. 

### Two Possible solutions (pick one or both) 

#### Supermarket only rules via tagging

So why ship Supermarket only rules with Foodcritic at all? Why not just create custom Supermarket rules we run locally there? Well I strongly believe that users should be able to run those same quality tests locally without significant effort. Requiring an external rules library is significant effort since we don't have a plugin sharing system in Foodcritic, and I have no interest in ever doing that. So what we need is a more formalized process for bringing those rules in and the understanding that we'll disable them by default. We also need to expose information on Supermarket showing how to run cookbooks locally with all these rules enabled. I'm curious what others think. I'd be open to creating disabled by default Foodcritic rules for anything that's been accepted to the cookbooks quality metrics repo first. 

#### "Warning" only rules and a better config system

So the second solution is to create a better config system and use that config system to implement different failure levels for rules. This would allow us to make some rules like this "warning" rules while others were "failure" rules. It's a pretty significant amount of work, but something worth biting off if we consider Foodcritic to be a tool we expect to use in 3-4 more years.

damacus:
Closing this until we have a better config system.

PR #677, opened 2017-10-17
Add FC097, FC098, FC099, and FC100 for deprecated mixin usage

tas50:

### DIFF
--- a/lib/foodcritic/rules/fc097.rb
+++ b/lib/foodcritic/rules/fc097.rb
@@ -1,0 +1,11 @@
+rule "FC097", "Deprecated Chef::Mixin::LanguageIncludeAttribute class used" do
+  tags %w{chef14 deprecated}
+  def lang_include_attrib_mixin(ast)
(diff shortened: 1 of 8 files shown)


coderanger:
Any reason to not make these all one rule?

tas50:
I considered making it one rule, but we'll have deprecation docs linked for these and the remediation for each is going to be different. Also if we ever deprecate a new mixin it would be less than cool to just add another one to the rule randomly. This makes it super explicit what we're checking and what failed. Also if the contents of the supermarket are any judge *very* few people are going to hit any of these. They're barely public APIs that weren't used much outside of Opscode.

coderanger:
I'm mostly just concerned about rule sprawl, but I suppose that ship already sailed :-/

tas50:
The rules are pretty fine grained, but IMO that's a feature not a bug. The tags are where the sprawl needs to be contained. You shouldn't be disabling individual rules you should be disabling tags

PR #683, opened 2017-10-19
Update FC087 to check for all deprecated Chef::Platform methods

tas50:
There's 3 deprecated methods and we deprecated them all at the same time. FC087 should have included them all from day 1


thommay:
Hm, I wonder if adding more stuff to 87 might not be a better plan?

tas50:
I get a bit nervous greatly expanding existing checks. I'd rather kill off the old ones and introduce new ones.

coderanger:
I'm with Thom on this one, making it a new rule seems more confusing given it's the same qualitative check for most people.

tas50:
@coderanger Fair enough. I updated FC087 for the 2 new methods

PR #690, opened 2017-11-08
Simplify how we define rules that need to run on all files

tas50:
Recipe is what we actually want. It includes it all.


tas50:
@coderanger Any objection to this cleanup.

PR #691, opened 2017-11-10
Add FC106: Use the gid property in user instead of group

tas50:
This is another alias in code where we've create two properties that do the same thing and only cause confusion. Point the users to the actual property.


coderanger:
Not sure I agree, that was added for semantic consistency. `gid 123` vs `group 'mygroup'`.

tas50:
Fair enough.

PR #693, opened 2017-11-10
Add FC104: Use the :run action in ruby_block instead of :create

tas50:
We have an alias here to prevent people from using :create. Most people on the supermarket are doing the right thing. The create action doesn't reflect what's actually happening with the resource and we should remove it from the docs entirely.


coderanger:
Do we want to also try to catch people sending `:create` as a notification? I think I've run in to it there too.

tas50:
@coderanger It should catch the notifies as well now even if a ruby_block itself does an improper notification

PR #695, opened 2017-11-10
Add FC108: Resource should not define a property named 'name'

tas50:
There's no need to redefine the name property in custom resources. It's all over the place in custom resource code and I see it in PRs for our cookbooks all the time. People think they have to do this.

### DIFF
--- a/lib/foodcritic/rules/fc108.rb
+++ b/lib/foodcritic/rules/fc108.rb
@@ -1,0 +1,12 @@
+rule "FC108", "Resource should not define a property named 'name'" do
+  tags %w{correctness}
+  resource do |ast|
+    # Make sure we're in a custom resource not an LWRP
+    if ast.xpath("//command/ident/@value='action'")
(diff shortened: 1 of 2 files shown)


tas50:
Hold off on this one it has some false alerts

tas50:
This is ready for review now

coderanger:
I feel like this needs some more tests, like `property :foo, default: :name`? This looks like it would fire on any property with `:name` anywhere in it. That said, it's super annoying to match only specific numbered arguments in our XPath sooooo :-/

coderanger:
I guess it's not too bad to match only the first argument because you can pretty easily check for the `'()` equiv.

tas50:
I can try giving the abs path to the child. It’ll be an ugly query but bullet proof

coderanger:
It's rare enough that it's probably not worth huge complications over, but if it ends up not being too hard, would be nice :)

tas50:
@coderanger I added a spec for that scenario and it works as expected.

PR #696, opened 2017-11-13
Fix FC086 to only alert when you try to load a databag

tas50:
Ignore the other methods in this class since there's valid uses for them within cookbook code


tas50:
The test failure is actually a line in the database cookbook that shouldn't be alerting. I'll update the regression tests to properly expect nothing here.

Issue #619, opened 2017-04-24
Update FC085 to detect @new_resource usage


tas50:
Right now we only fire on new_resource. We should fire on both new_resource and @new_resource as both are used extensively in community cookbooks

flaccid:
I'm already feeling this?
`FC085: Resource using new_resource.updated_by_last_action to converge resource: ./providers/environment.rb:29`

tas50:
There's 137 uses of @new_resource.updated_by_last_action on the supermarket so it's definitely worth finding

PR #697, opened 2017-11-14
Fix FC085 to alert if someone uses @new_resource.updated_by_last action

tas50:
There's 137 occurrences of this on Supermarket.


tas50:
Fixes #619

tas50:
@coderanger This should be ready now.

PR #698, opened 2017-11-14
Add FC106: Use the plist_hash property in user instead of hash

tas50:
We changed hash to plist_hash in chef 13


lib/foodcritic/rules/fc106.rb
```diff
@@ -0,0 +1,8 @@
+rule "FC106", "Use the plist_hash property in user instead of hash" do
```
coderanger: Do you mean `in launchd`?

tas50:
@coderanger Name fixed. The copy from the old rule is showing here

PR #703, opened 2017-12-01
FC009 should not alert on a raise in a resource

tas50:
It's not the cleanest code, but it's not FC009's business


tas50:
I'll be honest I felt the same way. It was a legit, but you probably shouldn't do that or run into it sorta situation. It was pretty easy to fix though so why not

PR #705, opened 2017-12-13
Don't require a space before the # in an ignore comment.

coderanger:
This is important for file-level ignores which go on line 1 for now. We should have a better comment system for file-level ignores but this is a stop-gap at least.


coderanger:
@tas50 Finally green. I r gud at code.

PR #706, opened 2018-01-03
Allow search for multiple resource types in the find_resources api method

tas50:
Type can still be a string, but it can also be an array of resource types. I was kinda surprised this wasn't already a thing here.

### DIFF
--- a/lib/foodcritic/api.rb
+++ b/lib/foodcritic/api.rb
@@ -191,7 +191,7 @@ module FoodCritic
       return [] unless ast.respond_to?(:xpath)
       scope_type = ""
       unless options[:type] == :any
-        type_array =  Array(options[:type]).map! { |x| "@value='#{x}'" }
+        type_array =  Array(options[:type]).map { |x| "@value='#{x}'" }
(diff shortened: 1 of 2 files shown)


lib/foodcritic/api.rb
```diff
@@ -179,17 +179,21 @@ def file_match(file)
...
+    #     find_resources(ast, type: 'service')
+    #     find_resources(ast, type: %w(service file))
     #
     def find_resources(ast, options = {})
       options = { type: :any }.merge!(options)
       return [] unless ast.respond_to?(:xpath)
       scope_type = ""
-      scope_type = "[@value='#{options[:type]}']" unless options[:type] == :any
+      unless options[:type] == :any
+        type_array =  Array(options[:type]).map! { |x| "@value='#{x}'" }
```
coderanger: You probably want `map` instead of `map!` here. If an array is passed in, this would try to actually mutate it (i.e. `Array(x)` doesn't copy if it's already an array).

tas50: it was map! when it made sense before I switched things up. Let me fix that up

PR #710, opened 2018-01-04
use_inline_resources is the default now. Deprecate FC017/057/059 and add FC113

tas50:
Chef 13 enabled use_inline_resources by default, which eliminates the need for several rules and means we need a new rule to check for the inclusion of use_inline_resources.


tas50:
This should go in a release of Foodcritic in ~March since Chef 12 goes EOL in April. We'll get it pulled into ChefDK 14.

tristanmorgan:
Sorry but is chef15 a typo?

tas50:
It will cause a failure in Chef 15 so it's tagged with chef15.

tristanmorgan:
Ohh, OK. I saw the "Chef 13 enabled use_inline_resources...", my mistake.

tas50:
Not needed with 13+, will warn in a future Chef 14 release, will fail in 15.0

PR #737, opened 2018-02-28
Corrects a typo in the tests for FC007

richardc:
This example was passing, but because it was not finding an
`include_recipe`, rather than finding a local use of `include_recipe`.

Signed-off-by: Richard Clamp <richardc@unixbeard.net>


tas50:
Nice catch. I'll see what I can do to fix this one up

richardc:
There shouldn't be anything to fix-up, the example when corrected still works, but now for the right reason.  

The travis failures are the same as those on master, problems cloning git repositories for the regression suite.

PR #742, opened 2018-03-07
Fixup failing travis on master

richardc:
Master is currently failing on travis:

* The latest release of rubocop adds new style checks, so the rubocop target was failing
* The recipe for removing other versions of bundler was not working under RVM 2.5.0, leading to two installs of bundler.

Here we:

* pins rubocop to the last known compatible release based on travis logs
* Adopts an approach from https://github.com/travis-ci/travis-ci/issues/8717#issuecomment-366500378 to 'uninstall' bundler from RVM 2.5.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ branches:
 
 # this avoids build issues with rainbow on old bundler/rubygems
 before_install:
-  #- gem update --system
+  - gem update --system
   - gem --version
   - rvm @global do gem uninstall bundler -a -x -I || true
   - gem install bundler
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ group :test do
   gem "minitest", "~> 5.3"
   gem "minitest-reporters"
   gem "simplecov", "~> 0.8"
-  gem "chefstyle", "~> 0.5"
+  gem "chefstyle", "~> 0.5.0"
 end
 
 group :development do


richardc:
So far just a pin to rubocop.

Issue with the `bundle exec` doesn't happen locally, will be pushing various investigation commits to see what travis environment is doing

richardc:
Regression tests are running in all cells, but only failing in the 2.5.0 rvm.  Hmm.

richardc:
To surface the error:

```
  1) regression test with cookbook users should match expected output
     Failure/Error: command("git clone -q https://github.com/chef-cookbooks/#{name}.git .")
     
     Mixlib::ShellOut::ShellCommandFailed:
       Expected process to exit with [0], but received '1'
       ---- Begin output of bundle exec git clone -q https://github.com/chef-cookbooks/users.git . ----
       STDOUT: 
       STDERR: /home/travis/.rvm/gems/ruby-2.5.0/bin/bundle:23:in `load': cannot load such file -- /home/travis/.rvm/rubies/ruby-2.5.0/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/exe/bundle (LoadError)
       	from /home/travis/.rvm/gems/ruby-2.5.0/bin/bundle:23:in `<main>'
       	from /home/travis/.rvm/gems/ruby-2.5.0/bin/ruby_executable_hooks:15:in `eval'
       	from /home/travis/.rvm/gems/ruby-2.5.0/bin/ruby_executable_hooks:15:in `<main>'
       ---- End output of bundle exec git clone -q https://github.com/chef-cookbooks/users.git . ----
       Ran bundle exec git clone -q https://github.com/chef-cookbooks/users.git . returned 1
     # ./vendor/bundle/ruby/2.5.0/gems/mixlib-shellout-2.3.2/lib/mixlib/shellout.rb:293:in `invalid!'
     # ./vendor/bundle/ruby/2.5.0/gems/mixlib-shellout-2.3.2/lib/mixlib/shellout.rb:280:in `error!'
     # ./vendor/bundle/ruby/2.5.0/gems/rspec-command-1.0.3/lib/rspec_command.rb:111:in `block in command'
     # ./vendor/bundle/ruby/2.5.0/gems/rspec-command-1.0.3/lib/rspec_command.rb:108:in `tap'
     # ./vendor/bundle/ruby/2.5.0/gems/rspec-command-1.0.3/lib/rspec_command.rb:108:in `command'
     # ./spec/regression/regression_spec.rb:11:in `block (4 levels) in <top (required)>'
     # ./vendor/bundle/ruby/2.5.0/gems/rspec-command-1.0.3/lib/rspec_command.rb:47:in `block (2 levels) in <module:RSpecCommand>'
     # ./vendor/bundle/ruby/2.5.0/gems/rspec-command-1.0.3/lib/rspec_command.rb:45:in `block in <module:RSpecCommand>'
```

So that's the `bundle exec git clone ...` failing to find `exe/bundle` from the gem path.

richardc:
https://travis-ci.org/Foodcritic/foodcritic/jobs/350259777#L491-L492

```
$ rvm @global do gem uninstall bundler -a -x -I || true
ERROR:  While executing gem ... (Gem::InstallError)
    gem "bundler" cannot be uninstalled because it is a default gem
```

Looks like the uninstall doesn't occur, so we end up with multiple installed/incorrectly activated bundler.

coderanger:
Deleting a default gem is not usually a safe thing to do. Why do you think we should do is here?

richardc:
@coderanger because otherwise you end up with two versions of bundler installed, which breaks as described in earlier comments https://github.com/Foodcritic/foodcritic/pull/742#issuecomment-371101183

If you have a fix for the bug in a better form I'm happy to defer to your fix

coderanger:
Does that still happen with an updated rubygems (i.e. uncomment the `gem update --system`)?

richardc:
Let's see - 9e6ff1e

tas50:
That looks to fix the issue. Can you squash this all down to a single commit. Also instead of pinning rubocop you can just set chefstyle = 0.5.0 which should do it. I'll get a PR open to cleanup the warnings there later. Thanks for taking this one though. It's been on my todo list since it started.

PR #745, opened 2018-04-04
Force UTF8 encoding

scotthain:
Signed-off-by: Scott Hain <shain@chef.io>

For example, when you're running in a container you don't control that doesn't have a utf8 locale set.

example:
```
LC_ALL=C foodcritic .
Checking 48 files
bundler: failed to load command: foodcritic (<snip>/.chefdk/gem/ruby/2.4.0/bin/foodcritic)
ArgumentError: invalid byte sequence in US-ASCII
```

### DIFF
--- a/lib/foodcritic.rb
+++ b/lib/foodcritic.rb
@@ -1,3 +1,6 @@
+Encoding.default_external = Encoding::UTF_8
+Encoding.default_internal = Encoding::UTF_8
+
 require "pathname"
 require "cucumber/core"
 require "treetop"


lamont-granquist:
prior art:  https://github.com/berkshelf/berkshelf/blob/1546390764b5c93a6f1845953032936b350f8878/lib/berkshelf.rb#L29

lamont-granquist:
you could probably move that into `lib/foodcritic.rb` if you wanted to use a larger hammer.

scotthain:
done @lamont-granquist

PR #748, opened 2018-04-07
Add FC116: Cookbook depends on the deprecated compat_resource cookbook

tas50:
compat_resource isn't required with later released of Chef 12 and should
not be used in modern cookbooks.


damacus:
Yes please.

/me runs this over sous-chefs.

PR #750, opened 2018-04-11
Add Chef 14.0.190 metadata and make it the default

tas50:
Users can still set things to 13.7 to get the 13.X behavior, but this makes FC aware of all the new resources


tas50:
This one is going to need a bit of work. It looks like we might have a regression in FC019 with the new Chef 14 data.

PR #753, opened 2018-04-11
Fix 2 bugs in calculating the cookbook dir from cookbook fles

tas50:
Both of these bugs happened when you matched on cookbook:

If you used the Chef 13+ recipe.rb or attribute.rb files we would end up 1 directory above the cookbook which would cause rules to fail
If you placed you templates directly in the templates directory you would end up 1 directory above the cookbook which would cause rules to fail

### DIFF
--- a/lib/foodcritic/linter.rb
+++ b/lib/foodcritic/linter.rb
@@ -264,8 +264,14 @@ module FoodCritic
       # we didn't find something in cache so look it up and cache it for later
       cook_val = Pathname.new(File.join(File.dirname(file),
                                         case File.basename(file)
-                                        when "metadata.rb" then ""
-                                        when /\.erb$/ then "../.."
+                                        when "recipe.rb", "attribute.rb", "metadata.rb" then ""
+                                        when /\.erb$/
+                                          if File.exist?(File.join(File.dirname(file), "../../metadata.rb")) ||
+                                             File.exist?(File.join(File.dirname(file), "../../metadata.json"))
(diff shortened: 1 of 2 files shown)


tas50:
This is differently broken:

FC071: Missing LICENSE file: chef-client/templates/default/LICENSE:1
FC071: Missing LICENSE file: chef-client/templates/default/clearlinux/LICENSE:1
FC071: Missing LICENSE file: chef-client/templates/default/debian/LICENSE:1
FC071: Missing LICENSE file: chef-client/templates/default/fedora/LICENSE:1
FC071: Missing LICENSE file: chef-client/templates/default/redhat/LICENSE:1
FC071: Missing LICENSE file: chef-client/templates/default/suse/LICENSE:1

lib/foodcritic/linter.rb
```diff
@@ -261,13 +261,30 @@ def cookbook_dir(file)
...
+      if file =~ /\.erb$/
+        # split each directory into an item in the array
+        dir_array = File.dirname(file).split(File::SEPARATOR)
+
+        # walk through the array of directories backwards until we hit the templates directory
+        position = -1
+        position -= 1 until dir_array[position] == "templates"
+
+        # go back 1 more position to get to the cookbook dir
+        position -= 1
```
damacus: This and 269 should/could be the same no?

tas50: if we're at say -3 already we need to make sure we're at -4. -3 would be the templates directory

PR #765, opened 2018-06-15
Bump ruby versions we test on in Travis

tas50:


tas50:
Same failure as before

PR #766, opened 2018-06-27
Remove FC012

tas50:
This is an outdated rule that looks for rdoc formatted readmes. No one uses these anymore and we already have a check that looks for readme.md so anyone using rdoc sees double alerts.


lamont-granquist:
ah, so old cookbooks on supermarket that use README.rdoc still get the no-README.md warning...  so ancient cookbooks that haven't been updated still throw something...

PR #767, opened 2018-06-27
Add FC121: Cookbook depends on cookbook made obsolete by Chef 14

tas50:
Check for users that are including cookbooks where the resources are
built into chef 14. I'm not tagging this with chef 14 since doing so
makes it seem like a thing you *must* fix to use chef 14. Getting rid of
these deps just speeds up chef and simplifies your depsolving a bit.

### DIFF
--- a/lib/foodcritic/rules/fc121.rb
+++ b/lib/foodcritic/rules/fc121.rb
@@ -1,6 +1,6 @@
 rule "FC121", "Cookbook depends on cookbook where resources are built into Chef 14" do
   tags %w{correctness}
   metadata do |ast, filename|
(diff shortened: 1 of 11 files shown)


lib/foodcritic/rules/fc121.rb
```diff
@@ -0,0 +1,6 @@
+rule "FC121", "Cookbook depends on cookbook where resources are built into Chef 14" do
```
lamont-granquist: The wording seems to read awkward, but I'm still sick-foggy-brain so not sure if that's me, and don't have a better idea.

tas50: I need @thommay

PR #772, opened 2018-06-28
Remove bogus resources from the metadata files

tas50:
Quick and dirty fix to remove a few more bogus resources. Slims the files down a lot.


lamont-granquist:
whyrun_safe_ruby_block is not bogus though?

PR #784, opened 2018-10-23
Disable FC121 until Chef 13 goes EOL

tas50:
I jumped the gun on this rule. We should wait until April.


damacus:
🎉

Issue #715, opened 2018-01-17
FC019 false positive on `node.force_default`


haidangwa:
This looks like a re-occurrence of #181 
---------------------------------------

chefdk: 2.4.17

My cookbook wraps the ssh_known_hosts cookbook. Foodcritic throws this FC019 violation exactly as discussed in #181.

My Gemfile:
```
gem 'berkshelf', '~> 6.3.1'
gem 'chef', '~> 12'
gem 'chef-zero'
gem 'chefspec', '~> 7.1.0'
gem 'foodcritic', '~> 12.2.2'
gem 'inspec' # constrained by kitchen-inspec
gem 'kitchen-ec2', '~> 1.4.0'
gem 'kitchen-inspec', '~> 0.20'
gem 'kitchen-sync', '~> 2.1'
gem 'rubocop', '~> 0.51.0'
gem 'test-kitchen', '~> 1.19.2'
```

`FC019: Access node attributes in a consistent manner`: The line that it complains about simply does
```
node.force_default['sshd']['config']['AuthorizedKeysFile'] = '/etc/ssh/authorized_keys/%u'
```

Changing this to use `node.default` instead will get rid of the violation warning.

TravisWhitehead:
I've researched this further. Even if I generate a `chef_dsl_metadata/chef_13.8.5.json` (beause I am using Chef 13.8.5) by following [these instructions](https://github.com/Foodcritic/foodcritic/blob/master/chef_dsl_metadata/README.md), this false-positive persists when I use `node.force_default`.

I noticed that in the generated file `force_default` is not present in the `dsl_methods` list, but if I add it to that list then no false-positive FC019 occurs.

In an old solution to #181, PR #364 [added `Chef::Node::Attribute.public_instance_methods` to `chef_node_methods` ](https://github.com/Foodcritic/foodcritic/pull/364/files#diff-d0a4237ec556e37293d4542e8fa30b58) as a solution to the FC019 false-positive discussed in #181.

I've tried adding `Chef::Node::Attribute.public_instance_methods` to `chef_dsl_methods` in [`chef_dsl_metadata/Rakefile`](https://github.com/Foodcritic/foodcritic/blob/master/chef_dsl_metadata/Rakefile#L21-L28) and have found that the files generated from this modified Rakefile no longer produce a false-positive FC019.

I'm unsure if this is a suitable solution, so I would like some input from the foodcritic community. I may make a PR soon requesting commentary on this fix.

PR #793, opened 2019-03-05
Fix node_method? to check chef_node_methods instead of chef_dsl_methods

TravisWhitehead:
**Edit:**
lamont-granquist pointed out that my original fix (described below) is wrong. The actual problem was that `node_method?` was checking `chef_dsl_methods` when it should have been `chef_node_methods`. This PR now fixes that method, removes an old workaround stemming from this bug from `chef_dsl_metadata/Rakefile`, and regenerates the `chef_dsl_metadata/` files.

-----

In #715 there is a false-positive FC019 occurring when using `node.force_default`. This PR proposes a solution that seems to prevent this false-positive, though based on the FIXME already present in the Rakefile I get the feeling that this may not be an ideal solution. This solution is inspired by the fix in #364 (which was addressing a similar bug in #181).

As I've discussed in #715, it seems that `force_default` must be present in the `dsl_methods` list in these generated `chef_dsl_metadata/chef_X.Y.Z.json` files, otherwise this false-positive FC019 occurs.

I would welcome some input from the Foodcritic community as to whether this is an appropriate fix, or if there's something better we can do.


lamont-granquist:
so #181 should have fixed this.   if there is a problem here it is that something is using `chef_dsl_methods` when it should be using `chef_node_methods`, which needs to get fixed.  this seems pretty obviously the wrong way to fix this.

lamont-granquist:
like why does `def node_method?` in lib/foodcritic/api use `chef_dsl_methods` and not `chef_node_methods`?

lamont-granquist:
yeah, as far as i can tell when this got written only chef_dsl_methods existed:

https://github.com/Foodcritic/foodcritic/commit/935a712cee7bc93b129109da1f0503ac04819caf

the node_dsl_methods were introduced later and FC019 was never updated to use it:

https://github.com/Foodcritic/foodcritic/commit/6a776d969d4b84373635f4e4e60b40991a7c2a1e

the mistake of including the node methods in the dsl methods dates back to when that was originally created and seems to be a straight up thinko as far as i can see:

https://github.com/Foodcritic/foodcritic/commit/1685d977948888f888d935fc7ec827bee4e04321

so `node_method?` should be changed to use `chef_node_methods`, the Chef::Node should actually be removed from `chef_dsl_methods`, and then any remaining newly exposed issue should get cleaned up.

TravisWhitehead:
Thank you for clarifying the source of the actual bug and the fix, as well as the quick response!

I've updated my branch with what I believe is the fix you described. It does resolve the false-positive that I was experiencing.

tas50:
Thanks @TravisWhitehead

PR #795, opened 2019-04-08
Remove Chef 13 metadata and add Chef 15

tas50:
Chef 13 will be EOL for the next DK release


chef_dsl_metadata/Appraisals
```diff
@@ -1,5 +1,4 @@
 chef_versions = [
-  "13.12.3",
   "14.10.9",
```
damacus: Worth adding a 15 at the same time?

tas50: We have to have a gem first

tas50: And now we have 15

PR #796, opened 2019-04-11
Read foodcritic rules from a url

MarkGibbons:
Allow the -r path to be a url. This change should make sharing
a common set of foodcritic rules easier.


lib/foodcritic/linter.rb
```diff
@@ -216,7 +217,6 @@ def rule_file_tags(file)
 
       # if a rule file has been specified use that. Otherwise use the .foodcritic file in the CB
       tags = if @options[:rule_file]
-               raise "ERROR: Could not find the specified rule file at #{@options[:rule_file]}" unless File.exist?(@options[:rule_file])
```
tas50: Is there a way we can still check for the location when we have a local file specified?

MarkGibbons: The read fails at line 241 if the file doesn't exist and returns
  raise "ERROR: Could not read or parse the specified rule file at #{file}"
I was hoping that was a good enough message about the file being missing.

 Otherwise I can try to figure out an unless that looks like:   unless something.is_local? && File.exist?

MarkGibbons: PR fix coming up.

PR #799, opened 2019-05-14
Test on Ruby 2.6 now

tas50:


.travis.yml
```diff
@@ -4,6 +4,7 @@ rvm:
   - 2.3.8
```
damacus: Do we still want to test on 2.3.8?

tas50: It's still a supported version so yep

PR #800, opened 2019-05-17
Allow configuring the size of the AST cache

Babar:
Also, make the cache unique, so if people want to use read_ast in a rule, it will use the same cache

Otherwise, it can get really slow if people check dependencies using AST. The default of 5 will keep the memory footprint low and should be fine for most people

### DIFF
--- a/lib/foodcritic/api.rb
+++ b/lib/foodcritic/api.rb
@@ -258,7 +258,7 @@ module FoodCritic
     # @since 16.1
     # @param size [Integer] the size of the cache (will be resized)
     def ast_cache(size = nil)
-      @@ast_cache ||= Rufus::Lru::Hash.new(size)
+      @@ast_cache ||= Rufus::Lru::Hash.new(size || 5)
       if size && @@ast_cache.maxsize != size
         @@ast_cache.maxsize = size
       end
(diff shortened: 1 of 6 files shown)


lib/foodcritic/api.rb
```diff
@@ -254,14 +254,16 @@ def match(node)
         line: pos["line"].to_i, column: pos["column"].to_i }
     end
 
+    def ast_cache(size = nil)
```
tas50: Can you add YARD comments here including the introduced field with the next feature release version of Foodcritic (16.1) so people know when this came about.

Babar:
The failing tests are due to using a single cache. I'll fix the tests.

jaymzh:
You still have test failures though...

Babar:
Pretty certain the last 2 failing tests are because I'm testing with `rbenv` and Travis CI runs with `rvm`. I honestly have no idea what's wrong, as it works for me, and even running manually what the test does works. (running foodcritic with no argument returns a non-zero exit code after printing the usage, same with running it with a non-existent -w option)

Babar:
@tas50 would you mind having a look please? Pretty sure the code is correct, but I have no idea how to debug as I don't use `rvm`

tas50:
Thanks @Babar. Sorry for the long delay here

